PR #1025, opened 2022-06-17
Fix/ Forum page: use correct dates in meta tags

zbialecki:
Previously `citation_publication_date` used the cdate of the note while `citation_online_date` used mdate. This PR reverses that.

Fixes https://github.com/openreview/openreview/issues/32

### DIFF
--- a/client/webfield-v2.js
+++ b/client/webfield-v2.js
@@ -338,6 +338,7 @@ module.exports = (function() {
       emptyMessage: 'No information to show at this time.',
       renders:[],
       headings: rows.length ? Object.keys(rows[0]) : [],
+      pageSize: null,
       extraClasses: '',
       reminderOptions: {
         container: 'a.send-reminder-link',
@@ -350,6 +351,8 @@ module.exports = (function() {
     options = _.defaults(options, defaults);
     var $container = $(container).empty();
     var containerId = container.slice(1);
+    var filteredRows = rows;
+    var selectedIds = [];
 
     var defaultRender = function(row) {
       var propertiesHtml = '';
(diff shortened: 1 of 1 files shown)


timokau:
I notice a similar issue is still present. As an example, see https://openreview.net/forum?id=rJe2syrtvS. The paper was published at ICLR 2020. The bibtex export correctly reflects that (year = 2020). The HTML page reports "Published: 20 Dec 2019", but the `citation_publication_date` field is set to `2019/09/25`, distinct from both the conference date and the camera-ready date.

@zbialecki, do you know what is going on here? Is this a bug, should I open a new issue?

melisabok:
This is what we got from Google Scholar: "citation_publication_date is the date the article was submitted. citation_online_date is that the article was last revised"

and that is the reason the publication date in openreview doesn't match with the citation_publication_date in Google Scholar, specially for those papers that were public before acceptance.

timokau:
I see. It seems confusing to me however that `citation_publication_date` does not match the bibtex export. This is particularly problematic since it is the date that is picked up by Zotero when saving an article using the browser extension. Zotero could implement a custom parser for openreview, but ideally this should not be necessary since purposes like this are what the meta tags are for right?

melisabok:
Well, we agree with you but it seems that Google Scholar has a different meaning for publication date.

timokau:
And it is important to stick to google scholar's definition to be indexed correctly?

melisabok:
the meta tag `citation_publication_date` was added so Google Scholar can index our papers, this meta tag was defined by them: https://scholar.google.com/intl/en/scholar/inclusion.html#indexing

Why Zotero using that tag? can Zotero read the `pdate` available in the OpenReview note object?

I think this issue applies to venues like ICLR where they become public before publication. All the other venues these dates should match.

timokau:
The page you [linked](https://scholar.google.com/intl/en/scholar/inclusion.html#indexing) states that `citation_publication_date` should contain "the date that would normally be cited in references to this paper from other papers". Doesn't that mean that `citation_publication_date` *should* match the bibtex export? Maybe the google scholar policy has changed?

As far as I know Zotero uses that tag because it is standardized and it can use a single [translator](https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js) for many websites. It would probably be possible to provide a custom openreview translator if necessary.

melisabok:
I can double check with Google Scholar again but the POC told us that citation_publication_date should be the date the paper becomes online. We were surprised too because we had to change our code. 

OpenReview creates the bibtex entry using the publication dates, AFAIK this is correct.

PR #1498, opened 2023-06-14
Fix/ Forum chat: add notifications

zbialecki:
- Add PWA features
- Add local notifications
- Prevent JS error caused by searching for excessively long search terms
- Fix various other bugs

Chat scroll behavior:

+ on page load: scroll to bottom
+ on layout change to chat: scroll to bottom
+ on filter change: scroll to bottom
+ on page load with noteId: scroll to chat note
+ on post new note: scroll to bottom
+ on receiving new note: scroll to bottom only if already at bottom of list
+ on click of reply note: scroll to note
+ on delete message: don't scroll

### DIFF
--- a/components/forum/FilterTabs.js
+++ b/components/forum/FilterTabs.js
@@ -1,6 +1,11 @@
 import Link from 'next/link'
 
-export default function FilterTabs({ forumId, forumViews, replyInvitations = [] }) {
+export default function FilterTabs({
+  forumId,
+  forumViews,
+  newMessageCounts,
+  replyInvitations = [],
+}) {
   if (!forumViews || !window) return null
 
   const currentHash = window.location.hash.slice(1)
@@ -19,6 +24,8 @@ export default function FilterTabs({ forumId, forumViews, replyInvitations = [] 
           return null
         }
 
+        const newMessageCount = newMessageCounts[primaryInvitation] ?? 0
+
         return (
           <li
             key={view.id}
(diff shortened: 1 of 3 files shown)


zbialecki:
A browser notification will be triggered when the user:

- is on the forum page and a new forum reply (of any type) is added. The exception to this is when they are in the Chat tab viewing the latest messages (aka scrolled to the bottom)
- is on the forum page and a new edit of the forum note is posted

We could add notifications for new messages (emails to the user that are shown on the notifications page) however this would require polling the /messages endpoint which might put too much load on the server.

xkopenreview:
some ideas, might not be practical

- when user is on any page and paper where user is an author of receive a reply
- when user is on any page and user get some assignments
- when pc is on any page and submission/review deadline passed for a venue that the pc is organizing
- when pc is on any page and a paper is reported

zbialecki:
I actually like all of those ideas Xukun, however until we implement a websocket API, the only way we can get those events from the server would be to poll the API endpoint. So this would imply polling multiple note endpoints, polling for invitations, plus more for the PCs. 

What could work is to create a separate /notifications API endpoint and a notifications collection that was indexed by profile id and cdate so it would be simple and fast to poll every few seconds. This would make the frontend and API less tightly coupled and make it easier to add new types of notifications in the future. It would probably make sense to do this even when we have a Websockets API in the future. @carlosmondra what do you think?

melisabok:
I wonder if we can merge this PR to keep testing it in the dev site.

melisabok:
@zbialecki I want to merge this today and keep working on the other issues in a new PR. Can we merge it?

PR #1716, opened 2023-11-02
Fix/ Use API 2 to import DBLP papers

melisabok:
it depends on 

- https://github.com/openreview/openreview-py/pull/1901
- https://github.com/openreview/openreview-api/pull/333
- https://github.com/openreview/openreview-py/pull/1993

### DIFF
--- a/components/webfield/ProgramChairConsole.js
+++ b/components/webfield/ProgramChairConsole.js
@@ -1,6 +1,7 @@
 /* globals promptError: false */
 import { useContext, useEffect, useState } from 'react'
 import { useRouter } from 'next/router'
+import groupBy from 'lodash/groupBy'
 import useUser from '../../hooks/useUser'
 import useQuery from '../../hooks/useQuery'
 import { Tab, TabList, TabPanel, TabPanels, Tabs } from '../Tabs'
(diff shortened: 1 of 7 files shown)


xkopenreview:
the following packages are missing from ubuntu 20.04 (image ubuntu-2004-focal-v20200701)

- libnss3
- libatk1.0-0
- libatk-bridge2.0-0
- libcairo2
- libcups2
- libgbm1
- libpango-1.0-0
- libpangocairo-1.0-0
- libx11-xcb1
- libxcb-dri3-0
- libxcomposite1
- libxcursor1
- libxdamage1
- libxkbcommon0
- libxrandr2
- libxrender1
- libxshmfence1
- libxss1

there are no missing packages based on 
[chromium docs](https://source.chromium.org/chromium/chromium/src/+/main:chrome/installer/linux/debian/dist_package_versions.json)
checked against ubuntu 22.04 image ubuntu-2204-jammy-v20220902
so it looks like there's no need to modify the startup script of template

components/DblpImportModal.js
```diff
@@ -209,9 +209,11 @@ export default function DblpImportModal({ profileId, profileNames, updateDBLPUrl
         }, 2000)
       }
     } catch (error) {
-      setMessage(
-        'An error occurred while importing your publications. Please try again later.'
-      )
+      const errorMessage =
+        error.name === 'TooManyError'
+          ? 'DBLP import quota has reached'
```
melisabok, line 214: what happens if the user gets this message? ask them to try again?

xkopenreview: i think this does not apply any more because rate limiting has been replaced with maxReplies
let me remove it

zbialecki (commented):
Is there a script to create the DBLP.org/-/Record invitation for testing locally?

zbialecki (approved):
Looks good 👍

melisabok:
I'm curious to know who added the abstract manually.

PR #1784, opened 2024-01-04
Feature/ Include client IP in server side API requests

zbialecki:
Add `X-Forwarded-From` HTTP header to server side API requests. Only requests that run on the Next.js server (where ctx.req is defined) will include this header.

The value of this header can include multiple comma separated IP addresses, see here for more: https://serverfault.com/questions/846489/can-x-forwarded-for-contain-multiple-ips

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing to OpenReview Web
+
+Welcome, and thank you for your interest in contributing to the OpenReview web interface!
+
+## Asking Questions
+
+The issue tracker is intended for bugs found in the OpenReview UI. If your question is related to how OpenReview works please first check our [documentation](https://docs.openreview.net/). If you cannot find the answer there, please contact the organizers of your venue or contact us at <info@openreview.net>.
+
+## Providing Feedback
+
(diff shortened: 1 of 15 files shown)


lib/api-client.js
```diff
@@ -152,22 +155,22 @@ const getById = async (apiPath, id, accessToken, apiVersion, queryParam) => {
...
-const getInvitationById = async (invitationId, accessToken, queryParam1, queryParam2) => {
-  const invitationObj = await getById('invitations', invitationId, accessToken, 2, queryParam1)
+const getInvitationById = async (invitationId, accessToken, queryParam1, queryParam2, remoteIp) => {
+  const invitationObj = await getById('invitations', invitationId, accessToken, 2, queryParam1, remoteIp)
   if (invitationObj) {
     return invitationObj
   }
 
-  return getById('invitations', invitationId, accessToken, 1, queryParam2 ?? queryParam1)
+  return getById('invitations', invitationId, accessToken, 1, queryParam1 ?? queryParam2, remoteIp)
```
xkopenreview, line 164: i think queryParam1 here is the v2 param
so it should check queryParam2 (v1 param) first

arr invitation edit page can be used for testing

zbialecki: I see what you mean. I changed this line to match the behavior of getNoteById, but I guess that was the one that had the bug.

xkopenreview: i see
please take a look at #1794

PR #1788, opened 2024-01-11
Fix/Some changes to feedback form and institution list UI

xkopenreview:
this pr should:
1. update feedback topic of adding institution to  "Please add to your list of publishing institutions"
2. to  show an X button so we can delete the whole text in order to select another reason
3. add the geolocation info when we add domains using the moderation UI

3 requires https://github.com/openreview/openreview-api/pull/331 so that state/province is allowed in api validation

### DIFF
--- a/components/FeedbackModal.js
+++ b/components/FeedbackModal.js
@@ -24,7 +24,7 @@ export default function FeedbackModal() {
   const profileSubject = 'My OpenReview profile'
   const submissionSubject = 'A conference I submitted to'
   const organizationSubject = 'A conference I organized'
-  const institutionSubject = 'Please add to your list of publishing institutions'
+  const institutionSubject = 'Please add my domain to your list of publishing institutions'
   const subjectOptions = [
     profileSubject,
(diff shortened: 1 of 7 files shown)


enrubio (commented):
I noticed when I resized the moderation page the institution table misaligns and it could be hard to figure out what box belongs to what field. Not sure how necessary it is to fix, what do you think @xkopenreview? Otherwise it looks good to me and everything works as expected.

Screenshot 2024-01-24 at 4 16 45 PM

xkopenreview:
> I noticed when I resized the moderation page the institution table misaligns and it could be hard to figure out what box belongs to what field. Not sure how necessary it is to fix, what do you think @xkopenreview? Otherwise it looks good to me and everything works as expected.
> 
> Screenshot 2024-01-24 at 4 16 45 PM

currently the whole moderation page is not responsive so i think we can leave it as it is

melisabok:
@enrubio can this PR be merged?

enrubio (approved):

PR #1794, opened 2024-01-18
Fix/ Some issues in api client and construct edit

xkopenreview:
this pr should fix the following issues:
1. getInvitationById  and getNoteById are passing v2 param to v1 query (https://github.com/openreview/openreview-web/pull/862/files#r930117068)
2. edit.note.content.{field}.readers checking is wrong. 
currently it checks for reader.const which was changed during development of api.
correct check is reader.param, if there's no reader.param it's const array value

### DIFF
--- a/client/view-v2.js
+++ b/client/view-v2.js
@@ -2340,7 +2340,7 @@ module.exports = (function () {
 
       var fieldReader = contentFieldValue.readers
       if (fieldReader) {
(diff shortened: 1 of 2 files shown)


zbialecki (approved):
I reviewed the uses of getInvitationById and getNoteById to make sure there were no other bugs and they all look correct.

melisabok (approved):

PR #1797, opened 2024-01-23
Fix/ Assignment page: show deploy link based on invitation configuration or note status

melisabok:
Some venues like Journals will run matchings for a subset of papers, we should allow them to deploy multiple configuration. Add this setting to the configuration invitation content.

Setting is set here: https://github.com/openreview/openreview-py/pull/1989

### DIFF
--- a/pages/assignments/index.js
+++ b/pages/assignments/index.js
@@ -64,7 +64,7 @@ const AssignmentRow = ({
   handleRunMatcher,
   handleDeployMatcher,
   referrer,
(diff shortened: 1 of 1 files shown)


zbialecki (approved):

xkopenreview:
is it a valid use case that the number of deployments is limited?

melisabok:
> is it a valid use case that the number of deployments is limited?

I haven't thought about that case.

xkopenreview (approved):

PR #1798, opened 2024-01-24
Fix/ Search page: sort results by date modified

zbialecki:
Show most recently updated notes first in results.

Related to https://github.com/openreview/openreview/issues/149

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -134,10 +134,11 @@ const Search = ({ appContext }) => {
           content: query.content || 'all',
           group: query.group || 'all',
           source: Object.keys(sourceOptions).includes(query.source) ? query.source : 'all',
+          sort: 'tmdate:desc',
           limit: 1000,
         },
         null,
-        { accessToken, resultsKey: 'notes', sort: 'none' }
+        { accessToken, resultsKey: 'notes', sort: 'tmdate:desc' }
       )
 
       if (searchRes.notes) {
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -134,10 +134,11 @@ const Search = ({ appContext }) => {
           content: query.content || 'all',
           group: query.group || 'all',
           source: Object.keys(sourceOptions).includes(query.source) ? query.source : 'all',
+          sort: 'tmdate:desc',
           limit: 1000,
         },
         null,
-        { accessToken, resultsKey: 'notes', sort: 'none' }
+        { accessToken, resultsKey: 'notes', sort: 'tmdate:desc' }
```
melisabok, line 141: I wish we can sort them by pdate too, is there a way to decide this in the webfield component?

zbialecki: This change is for the general /search page, not the venue homepage search.

melisabok (approved):

pages/search.js
```diff
@@ -134,10 +134,11 @@ const Search = ({ appContext }) => {
           content: query.content || 'all',
           group: query.group || 'all',
           source: Object.keys(sourceOptions).includes(query.source) ? query.source : 'all',
+          sort: 'tmdate:desc',
```
xkopenreview, line 137: i think only of sort here and sort in options is required?
it's taking options.sort or data1.sort

zbialecki: Ah you are correct, I can remove it from the options object.

xkopenreview (approved):

PR #1800, opened 2024-01-29
Feature/Profile edit page - validate current history

xkopenreview:
discussed in the daily meeting to have a validation that a profile must have current (no end date or end date is current year or in future) education/career info


components/profile/ProfileEditor.js
```diff
@@ -197,6 +197,13 @@ export default function ProfileEditor({
         'You must enter position, institution, and domain information for each entry in your education and career history'
       )
     }
+    if (!profileContent.history.some((p) => !p.end || p.end >= new Date().getFullYear())) {
+      return promptInvalidValue(
+        'history',
+        profile.history?.[0]?.key,
+        'You must enter current education and career info'
```
melisabok, line 204: I think we should have a message less confusing because some users might not be studying. 

What about? "You must enter at least one current history record that could be education or career info" @celestemartinez please help us with the wording :)

components/profile/ProfileEditor.js
```diff
@@ -197,6 +197,13 @@ export default function ProfileEditor({
         'You must enter position, institution, and domain information for each entry in your education and career history'
       )
     }
+    if (!profileContent.history.some((p) => !p.end || p.end >= new Date().getFullYear())) {
+      return promptInvalidValue(
+        'history',
+        profile.history?.[0]?.key,
+        'You must enter at least one current history record that could be education or career info'
```
melisabok, line 204: it seems there is another error that says "Error: You must enter at least one position for your education and career history", should we change this too?

xkopenreview: education and career history cannot be empty?

xkopenreview:
require fix of https://github.com/openreview/openreview-api/pull/344#discussion_r1474797722 for test to pass

melisabok (approved):

components/profile/ProfileEditor.js
zbialecki: I feel like this error message doesn't convey that it needs to be a current position. I would say, "Your Education & Career History must include at least one current position."

celestemartinez (approved):

PR #1802, opened 2024-01-29
Fix/ Invitation editor - date process start/end date pass delete:true

xkopenreview:
currently invitation editor pass null when a date process cron job has no startDate and endDate
which is not allowed by invitation edit schema.

this pr should change it to pass {delete:true} instead


melisabok (approved):

carlosmondra:
Needs: https://github.com/openreview/openreview-api/pull/343

melisabok:
@xkopenreview is this PR ready?

zbialecki (approved):

PR #1805, opened 2024-01-30
Feature/Tasks - remove replyto filtering

xkopenreview:
this pr should remove replyto filtering of tasks

it should show link to invitation if it has webfield instead


xkopenreview:
this is required to show ICML author ranking invitation in author tasks
it has webfield (author should click the task to go to invitation page) but does not have replyto or note id

melisabok (approved):

PR #1809, opened 2024-02-02
Fix/ Author console - show confidence mark only

xkopenreview:
Emilia mentioned in the daily meeting that author console should show only the confidence number

this pr should update the author console to show only the confidence number


melisabok (approved):

enrubio (approved):
Looks good, thanks Xukun!

PR #1810, opened 2024-02-02
Fix/ Moderation - update v2 config note

xkopenreview:
currently moderation page is reading the moderate status and terms timestamp from v1 config note and is also updating only the v1 config note when moderate value is updated.

this pr should change it to read the value from v2 config note instead
changing moderate status will update v2 config note because new profiles are now created calling v2 api
changing terms timestamp will still update both config notes

### DIFF
--- a/pages/user/moderation.js
+++ b/pages/user/moderation.js
@@ -33,7 +33,7 @@ const UserModerationTab = ({ accessToken }) => {
   const [configNote, setConfigNote] = useState(null)
   const [configNoteV2, setConfigNoteV2] = useState(null)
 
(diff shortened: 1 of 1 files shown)


melisabok (approved):

racheljsmart:
I'm trying this on local. When set to Disabled the API 2 note has the moderate value set to No, and the API 1 note has the moderate value set to Yes. When enabled again, API 2 note has moderate set to Yes and API 1 has moderate set to Yes.

This is expected?

racheljsmart (approved):

zbialecki (approved):

PR #1812, opened 2024-02-06
Fix/Recruitment - Add captcha

xkopenreview:
there are still some users reporting auto accepting/declining during recruitment

this pr should add a captcha to the recruitment form component and have the accept/reject buttons disabled until the validation pass

this pr should be updated to pass the token to api when api require token to post a response for recruitment


carlosmondra:
I still don't know how the API will read the token from turnstile. Somehow the client needs to pass the token. I suggested passing the token in the headers, but I'm not sure if this is the best idea. Also, we need to figure out when the token is necessary. Should this be specified in the invitation?

xkopenreview:
> I still don't know how the API will read the token from turnstile. Somehow the client needs to pass the token. I suggested passing the token in the headers, but I'm not sure if this is the best idea. Also, we need to figure out when the token is necessary. Should this be specified in the invitation?

currently in this pr the token is only used to prevent bot from clicking the button on behalf of users. it's not passed to api with the accept/reject note

melisabok:
so we can use the captcha without making changes to the API?

xkopenreview:
> so we can use the captcha without making changes to the API?

yes.
this pr won't protect the end points that guest can post to though. this pr should solve the issue of bot clicking buttons on behalf of user

carlosmondra:
If this is sufficient to avoid bots from clicking buttons, then, no changes are needed in the API

melisabok:
Do openreview-py tests work with this change?

carlosmondra:
> Do openreview-py tests work with this change?

If they don't, I think there is a way to make them work: https://developers.cloudflare.com/turnstile/reference/testing/
I believe Xukun knows about this.

xkopenreview:
> Do openreview-py tests work with this change?

I tried to run all py tests but it keep getting stuck at waiting for some jobs to complete.
is there any particular test you are worried about?

carlosmondra:
> > Do openreview-py tests work with this change?
> 
> I tried to run all py tests but it keep getting stuck at waiting for some jobs to complete. is there any particular test you are worried about?
> 
> as @carlosmondra mentioned the token we used is always pass so even if it fail the fix will be just adding some wait time

Just make sure that a test that clicks on the Accept invitation to become a reviewer works. That's the only kind of test that could fail with this change.

melisabok:
> is there any particular test you are worried about?

there are severa tests that test the recruitment feature using selenium. 

One example is test_icml_conference.py

xkopenreview:
> > is there any particular test you are worried about?
> 
> there are severa tests that test the recruitment feature using selenium.
> 
> One example is test_icml_conference.py

it passed in my local

melisabok:
is it possible to put the widget centered and below the accept and decline buttons?

Screen Shot 2024-02-09 at 12 05 23 PM

does it say "testing only" because we are using localhost?

xkopenreview:
> is it possible to put the widget centered and below the accept and decline buttons?

please take a look at [8d006a0](https://github.com/openreview/openreview-web/pull/1812/commits/8d006a0724af41db1c41a099a7ac709af2a1b9ec)
> does it say "testing only" because we are using localhost?

yes. because the sitekey in local(env.example) is an always pass test key

racheljsmart (approved):

melisabok (approved):

melisabok:
@zbialecki do you want to take a final look?

PR #1813, opened 2024-02-09
Feature/ AC Console - support secondary AC in triplet AC assignments tab

xkopenreview:
this pr should add support of secondary ac assignments in ac console triplet ac assignments tab

example config:
>secondaryAreaChairName: "Secondary_Area_Chairs"

menu bar is not included in this pr due to time. it requires separating tab data from assigned papers tab because the "tableRowsAll" name in setData is fixed

### DIFF
--- a/components/webfield/AreaChairConsole.js
+++ b/components/webfield/AreaChairConsole.js
@@ -599,7 +599,7 @@ const AreaChairConsole = ({ appContext }) => {
               key={row.note.id}
               rowData={row}
               venueId={venueId}
-              areaChairName={secondaryAreaChairName}
+              areaChairName={areaChairName}
               officialReviewName={officialReviewName}
               officialMetaReviewName={officialMetaReviewName}
               submissionName={submissionName}
--- a/components/webfield/AreaChairConsole.js
+++ b/components/webfield/AreaChairConsole.js
@@ -699,7 +699,7 @@ const AreaChairConsole = ({ appContext }) => {
               active={activeTabId === '#triplet-ac-assignments' ? true : undefined}
               onClick={() => setActiveTabId('#triplet-ac-assignments')}
             >
-              Triplet AC Assignments
+              Secondary AC Assignments
             </Tab>
           )}
           <Tab
--- a/components/webfield/AreaChairConsole.js
+++ b/components/webfield/AreaChairConsole.js
@@ -204,7 +204,7 @@ const AreaChairConsole = ({ appContext }) => {
       )
       const areaChairGroups = allGroups.filter((p) => p.id.endsWith(areaChairName))
       const secondaryAreaChairGroups = secondaryAreaChairName
-        ? allGroups.filter((p) => p.id.includes(`/${secondaryAreaChairName}_`))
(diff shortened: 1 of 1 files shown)


melisabok:
Thanks Xukun, checking this now. @celestemartinez if you have to have these groups created in your local database feel free to test it.

I will extend the cvpr test to include seconday area chairs

components/webfield/AreaChairConsole.js
```diff
@@ -199,6 +203,9 @@ const AreaChairConsole = ({ appContext }) => {
         { accessToken }
       )
       const areaChairGroups = allGroups.filter((p) => p.id.endsWith(areaChairName))
+      const secondaryAreaChairGroups = secondaryAreaChairName
+        ? allGroups.filter((p) => p.id.includes(`/${secondaryAreaChairName}_`))
```
melisabok, line 207: I think we need to pass `Secondary_Area_Chairs` name instead of the singular version. 

A user could be added to the group but then removed and the anon group is not deleted. The source of truth should be VENUE_ID/Submission1/-/Secondary_Area_Chairs

melisabok: line 205 should be `const areaChairGroups = allGroups.filter((p) => p.id.endsWith('/' + areaChairName))`

and 206 can be `const secondaryAreaChairGroups = allGroups.filter((p) => p.id.endsWith('/' + secondaryAreaChairName))`

xkopenreview: i think / is not necessary? because it's checking endsWith

melisabok: Area_Chairs will match with Secondary_Area_Chairs, do we want that?

components/webfield/AreaChairConsole.js
```diff
@@ -624,18 +685,38 @@ const AreaChairConsole = ({ appContext }) => {
...
             Assigned Papers
           </Tab>
-          <Tab id="areachair-tasks" onClick={() => setActiveTabIndex(1)}>
+          {secondaryAreaChairName && (
+            <Tab
+              id="triplet-ac-assignments"
+              active={activeTabId === '#triplet-ac-assignments' ? true : undefined}
+              onClick={() => setActiveTabId('#triplet-ac-assignments')}
+            >
+              Triplet AC Assignments
```
melisabok, line 701: @celestemartinez should we call this "Secondary AC Assignments" so it is more general?

celestemartinez: yes, only CVPR has triplets but other venues like 3DV have used a single secondary AC

celestemartinez:
Tested this in my local and it looks good!

components/webfield/AreaChairConsole.js
```diff
@@ -624,18 +686,38 @@ const AreaChairConsole = ({ appContext }) => {
...
```
melisabok, line 702:
```suggestion
              Secondary AC Assignments
```

melisabok (approved):

celestemartinez (approved):

PR #1815, opened 2024-02-09
Fix/ Add Secondary AC or AC in missing readers

xkopenreview:
currently when secondary ac post a note when secondary ac group is not in note readers, the secondary ac won't be added as note readers

this pr should fix this issue by adding secondary ac group or ac group in note readers
this pr assumes that secondary ac group is member of ac group


melisabok (approved):
🔢

PR #1829, opened 2024-02-20
Fix/Reviewer Console - allow review display field names to be specified

xkopenreview:
Currently reviewer console always show the "review" field of a review note and it's not always a valid field

this pr should add in reviewer config a 
> reviewDisplayFields

Array field which has default value ["review"] and fields defined in the array will be shown (if review do have the field)


zbialecki:
I'm not sure why venues need to be changing the field name when we have the ability to change the display name in `presentation`. I would much rather we use that ability than parameterize this content field name in a number of different places.

melisabok:
Organizers decide their own review fields and they don't usually use one single field to enter the review text. Lately they have been using several fields, I would recommend the console support "display fields" as a list of field that will be shown in the reviewers console when the review is available. Is it possible that @xkopenreview ?

zbialecki:
OK, if that's the approach you want to take I would like to establish a convention for naming these config vars in all the consoles that need them. Maybe something like `reviewDisplayFields` and `metaReviewDisplayFields`?

melisabok:
it looks good to me

zbialecki (approved):

melisabok (approved):

celestemartinez (approved):

PR #1830, opened 2024-02-20
Feature/PC console - Add SAC to query search

xkopenreview:
This pr should fix the following 2 issues:
1. search value is converted to lowercase so "==" in query search may not work
2. to add sac to query search in pc console


zbialecki (commented):
Is it required that we have separate filters for case sensitive and case insensitive search? I can't think of a case where case sensitive search is required.

unitTests/webfield-utils.test.js
```diff
@@ -523,6 +523,70 @@ describe('filterCollections', () => {
...
+      collections,
+      filterString,
+      filterOperators,
+      propertiesAllowed,
+      uniqueIdentifier
+    )
+    expect(result.filteredRows.map((p) => p.id)).toEqual([2, 3])
+
+    // exact match
+    filterString = 'sac=="Name two"'
```
melisabok, line 569: Can we filter by tilde id too?

xkopenreview:
> Is it required that we have separate filters for case sensitive and case insensitive search? I can't think of a case where case sensitive search is required.

= is includes search
while == is for exact search so i think the filtering should not do any case conversion, it should compare with the exact string given by the user

unitTests/webfield-utils.test.js
xkopenreview: yes, added in [98e23a1](https://github.com/openreview/openreview-web/pull/1830/commits/98e23a19285461925879021875bd46a4828576fc)

zbialecki:
> = is includes search

OK, that makes sense.

zbialecki (approved):

melisabok (approved):

PR #1833, opened 2024-02-22
Fix/ Add support for floats to parseNumberField

haroldrubio:
This PR uses a regex to match the first number of a field and parse it as either a float or an integer

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1608,17 +1608,13 @@ export function getVenueTabCountMessage(unrepliedCommentCount, notDeployedCount)
 
 export function parseNumberField(value) {
   if (typeof value === 'string') {
-    const regex = /\d*\.?\d+/;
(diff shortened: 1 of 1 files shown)


zbialecki:
Where is this used? Can you give some steps to test it?

haroldrubio:
You can go to the EACL 2024 PC console on the live site to see some examples and compare them to this branch, here is an example of a before and after:
beforeChange
afterChange

melisabok:
How is the value stored in the note?

haroldrubio:
I see, the EACL data was posted using the old enum format using only strings, it looks like data with type float using an enum of objects renders correctly. We can close this PR in that case

zbialecki:
Yes, let's go with using an enum of objects then. Thanks for checking that.

PR #1840, opened 2024-03-01
Fix/Temp fix of PC console for venue with senior reviewers

xkopenreview:
this pr should add a component
>ProgramChairWithSeniorReviewerConsole

as temp fix for venue with senior reviewers until multiple reviewer groups is supported
(ProgramChairWithSeniorReviewerConsole.js can be deleted after the conference with senior reviewer is over)

this component has the following extra parameters compared to pc console (CoLLAs 2024 as an example):
```javascript
...
seniorReviewersId :'lifelong-ml.cc/CoLLAs/2024/Conference/Senior_Reviewers',
seniorOfficialReviewName: 'Senior_Official_Review',
seniorReviewerName: 'Senior_Reviewers',
anonSeniorReviewerName: 'Senior_Reviewer_',
...
```

function is the same as pc console with added info for senior reviewers in overview tab,paper status tab and senior reviewer status tab

decided that messaging senior reviewers from paper status tab is not supported for now
also assuming the reviewRatingNames are the same for reviewers and seniorReviewers

### DIFF
--- a/components/webfield/ProgramChairConsole/Overview.js
+++ b/components/webfield/ProgramChairConsole/Overview.js
@@ -1055,4 +1055,14 @@ const Overview = ({ pcConsoleData }) => {
 }
 
 export default Overview
-export { StatContainer }
+export {
+  StatContainer,
+  RecruitmentStatsRow,
+  SubmissionsStatsRow,
+  BiddingStatsRow,
+  MetaReviewStatsRow,
+  CustomStageStatsRow,
+  DecisionStatsRow,
+  DescriptionTimelineOtherConfigRow,
+  renderStat,
+}
--- a/components/webfield/ProgramChairWithSeniorReviewerConsole.js
+++ b/components/webfield/ProgramChairWithSeniorReviewerConsole.js
@@ -1,0 +1,2298 @@
+/* globals $,promptError: false */
+import { useContext, useEffect, useState } from 'react'
+import { useRouter } from 'next/router'
+import { camelCase, upperFirst } from 'lodash'
+import useUser from '../../hooks/useUser'
+import useQuery from '../../hooks/useQuery'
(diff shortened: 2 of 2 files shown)


melisabok:
Instead of senior reviewer can we call it alternate reviewer? There could be other venues that use different names for the second reviewer group. I'm open to use other names.

xkopenreview:
> Instead of senior reviewer can we call it alternate reviewer? There could be other venues that use different names for the second reviewer group. I'm open to use other names.


seniorOfficialReviewName and seniorReviewerName are the variable holding the value of "senior reviewers" or "alternate reviewers" or something else

their value can be anything

xkopenreview:
#1839 needs to be merged for test to pass

melisabok:
@racheljsmart can you test this in your local environment? I believe the component must be called:

```
component: 'ProgramChairWithSeniorReviewerConsole.js]'
```

and pass the properties that Xukun is describing, is that right @xkopenreview ?

xkopenreview:
> @racheljsmart can you test this in your local environment? I believe the component must be called:
> 
> ```
> component: 'ProgramChairWithSeniorReviewerConsole.js]'
> ```
> 
> and pass the properties that Xukun is describing, is that right @xkopenreview ?

just the name of the component will do:
current pc console:
>component: 'ProgramChairConsole'

to use the pc console with senior reviewers:
>component: 'ProgramChairWithSeniorReviewerConsole'

racheljsmart (approved):
I think this looks great! I tested on Collas and all the tabs look good and the progress sections count the notes and invitations correctly from what I could test.
I tried this for AutoML, they are not as far along as Collas so there was limited information, but it also seemed to work as needed for this group.

racheljsmart (approved):

PR #1844, opened 2024-03-06
Fix/Group editor: load all related invitations

xkopenreview:
this pr should make group editor related invitations section to use a regex prefix that can load more invitations such as 
>thecvf.com/CVPR/2024/Conference/Submission4/Meta_Review1/-/Final_Revision 

where "-" does not follow group id directly (user is at thecvf.com/CVPR/2024/Conference/Submission4 group)

### DIFF
--- a/components/group/GroupEditor.js
+++ b/components/group/GroupEditor.js
@@ -39,7 +39,7 @@ const GroupEditor = ({ group, isSuperUser, profileId, accessToken, reloadGroup }
       )}
       <GroupSignedNotes groupId={group.id} accessToken={accessToken} />
       <GroupChildGroups groupId={group.id} accessToken={accessToken} />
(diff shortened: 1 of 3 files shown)


melisabok:
Checking the network calls, the group editor is doing get invitations using this regex `ICML.cc/2023/Conference/Submission1/-/.*` to the API1, is it possible to avoid this call given the group is in API 2?

melisabok:
There is also a call to get notes using API 1, can this be removed too?

zbialecki:
Currently the code is using `getCombined` to load invitations from both API 1 and 2, but I guess it would make more sense to only make 1 call depending on whether the group is an API 2 group or not.

melisabok:
Ok, I will merge this PR and create an issue to remove these calls.

melisabok (approved):

PR #1847, opened 2024-03-07
Fix/ Don't allow user to import from DBLP with coauthor's persistent URL

xkopenreview:
currently a use can use a dblp url as long as any of the user's names appear in any of the listed papers
this pr should only allow import if the user's name match with the names specified in dblp xml

this is part of the preparation for auto profile merge


zbialecki (approved):
Works as expected

melisabok (approved):

PR #1853, opened 2024-03-08
Fix/ Forum page: Google Scholar changes

### LINKED ISSUE #1850: Forum pages should return a 403 error instead of redirecting to login when a Google crawler is dected
Currently, when a logged out user tries to access a non-public forum page, openreview-web redirects the user to the login page. In order to indicate to Google Scholar that this page should not be included in the index, we should instead return a HTTP 403 error when the Google crawler user agent string is detected.

### LINKED ISSUE #1851: Forum pages shouldn't render citation_published_date meta tags when citation_online_date is present
When a forum note includes an odate field only the citation_online_date meta tag should be used to prevent any confusion for Google Scholar.

zbialecki:
Fixes #1850 
Fixes #1851


pages/forum.js
```diff
@@ -100,8 +99,11 @@ const ForumPage = ({ forumNote, query, appContext }) => {
             {authors.map((author) => (
               <meta key={author} name="citation_author" content={author} />
             ))}
-            <meta name="citation_publication_date" content={onlineDate} />
-            <meta name="citation_online_date" content={modificationDate} />
+            {onlineDate ? (
+              <meta name="citation_online_date" content={onlineDate} />
+            ) : (
+              <meta name="citation_publication_date" content={publicationDate} />
+            )}
```
zbialecki, line 106: The logic here is basically: if the note has `odate` set use that for citation_online_date and don't include the citation_publication_date tag. Otherwise, show the citation_publication_date tag with the `pdate` or `cdate` of the note.

xkopenreview (approved):
looks good

pages/forum.js
melisabok: I think there shouldn't be public note with no odate

melisabok: my only concern about removing the tag citation_publication_date is that some other tools may use it. For example: zotero (https://www.zotero.org/support/dev/exposing_metadata)

Looking at the documentation:

```
Date: citation_date (preferred), citation_publication_date, citation_cover_date, citation_online_date, citation_year
```

I think having citation_online_date only is enough.

zbialecki: So all of the older venues are have odate? I thought some of them might be using pdate or not have either.

melisabok (approved):

pages/forum.js
melisabok: I think @carlosmondra set the odate for all of them.

PR #1854, opened 2024-03-11
Feature/PC Console with Senior Reviewers - add email senior reviewers options

xkopenreview:
this pr should add emailing senior reviewers and emailing reviewers + senior reviewers function in ProgramChairWithSeniorReviewerConsole paper status tab as requested by CoLLAs 2024 organizers

### DIFF
--- a/components/webfield/ProgramChairWithSeniorReviewerConsole.js
+++ b/components/webfield/ProgramChairWithSeniorReviewerConsole.js
@@ -1,7 +1,8 @@
-/* globals $,promptError: false */
+/* globals $,promptError,promptMessage: false */
 import { useContext, useEffect, useState } from 'react'
 import { useRouter } from 'next/router'
-import { camelCase, upperFirst } from 'lodash'
+import { camelCase, uniqBy, upperFirst } from 'lodash'
+import List from 'rc-virtual-list'
 import useUser from '../../hooks/useUser'
 import useQuery from '../../hooks/useQuery'
 import { Tab, TabList, TabPanel, TabPanels, Tabs } from '../Tabs'
@@ -34,7 +35,6 @@ import SeniorAreaChairStatus from './ProgramChairConsole/SeniorAreaChairStatus'
 import ReviewerStatusTab from './ProgramChairConsole/ReviewerStatus'
 import ErrorDisplay from '../ErrorDisplay'
 import RejectedWithdrawnPapers from './ProgramChairConsole/RejectedWithdrawnPapers'
-import MessageReviewersModal from './MessageReviewersModal'
 import LoadingSpinner from '../LoadingSpinner'
 import Table from '../Table'
 import NoteSummary from './NoteSummary'
@@ -44,6 +44,7 @@ import PaginationLinks from '../PaginationLinks'
 import Collapse from '../Collapse'
 import BaseMenuBar from './BaseMenuBar'
 import QuerySearchInfoModal from './QuerySearchInfoModal'
+import BasicModal from '../BasicModal'
 
 const SelectAllCheckBox = ({ selectedNoteIds, setSelectedNoteIds, allNoteIds }) => {
   const allNotesSelected = selectedNoteIds.length === allNoteIds?.length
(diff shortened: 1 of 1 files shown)


xkopenreview:
> Is this webfield currently only used by CoLLA 2024?

yes, there's another one which will be using it (autoML?) then it should be replaced by the function to support multiple reviewer groups

components/webfield/ProgramChairWithSeniorReviewerConsole.js
```diff
@@ -455,6 +456,279 @@ const Overview = ({ pcConsoleData }) => {
...
+          row.seniorReviewers.map((reviewer) => ({ ...reviewer, isSeniorReviewer: true }))
+        )
+      case 'allReviewersAndSeniorReviewers':
+        return selectedRows.flatMap((row) => [
+          ...row.reviewers.map((reviewer) => ({ ...reviewer, isSeniorReviewer: false })),
+          ...row.seniorReviewers.map((reviewer) => ({ ...reviewer, isSeniorReviewer: true })),
+        ])
+      case 'withReviews':
+        return selectedRows
+          .flatMap((row) => row.reviewers)
```
zbialecki, line 575: It seems inefficient to have 2 sequential calls to flatMap, when you could just combine them into 1 function, right?

components/webfield/ProgramChairWithSeniorReviewerConsole.js
```diff
@@ -455,6 +456,279 @@ const Overview = ({ pcConsoleData }) => {
   )
 }
 
+const MessageReviewersModal = ({
```
zbialecki, line 459: It seems like there's a lot of shared code between this and the MessageReviewersModal and the component used by the other consoles... if this is going to continue to be used by venues I wonder if there's a way to combine them. Possibly the MessageReviewersModal could have a param that takes a function that sends the emails, so you could just pass the function that starts on line 502 to the component.

components/webfield/ProgramChairWithSeniorReviewerConsole.js
xkopenreview: i think it's the same?
the 2nd flatmap is used here as filter by hasReview + map to add isSeniorReviewer
so it will become nested flatmap if combined into 1 function

components/webfield/ProgramChairWithSeniorReviewerConsole.js
xkopenreview: MessageReviewersModal is copied inside this console for easier cleanup so when these 2 conferences are over or when multiple reviewer groups are added in the venue request form only this console file need to be deleted.

i think message reviewers function only need to handle multiple reviewer groups, the email sending function should be the same.

components/webfield/ProgramChairWithSeniorReviewerConsole.js
zbialecki: Yeah a nested map wouldn't really be better

components/webfield/ProgramChairWithSeniorReviewerConsole.js
zbialecki: OK, so you are saying this feature of multiple reviewers groups will possibly be added to all the consoles and the  venue request form in the future.

I guess it makes sense to wait to modify MessageReviewersModal until then

zbialecki (approved):

components/webfield/ProgramChairWithSeniorReviewerConsole.js
xkopenreview: correct, reviewer groups will be an option in request form so all consoles will have it

racheljsmart:
> > Is this webfield currently only used by CoLLA 2024?
> 
> yes, there's another one which will be using it (autoML?) then it should be replaced by the function to support multiple reviewer groups

As of the 12th both AutoML/2024/Conference and AutoML/2024/ABCD_Track are using the ProgramChairWithSeniorReviewerConsole webfield

racheljsmart (approved):

PR #1857, opened 2024-03-13
Fix/Group Editor: remove calls to API 1 when the group is created in API 2

xkopenreview:
currently group editor use getCombined to load signed notes and related invitations.
this pr should change it to only load from api which match with the group version by checking group.domain

### DIFF
--- a/components/group/GroupRelatedInvitations.js
+++ b/components/group/GroupRelatedInvitations.js
@@ -17,25 +17,26 @@ const GroupRelatedInvitations = ({ group, accessToken }) => {
   const [totalCount, setTotalCount] = useState(null)
 
   const loadRelatedInvitations = async (limit, offset) => {
-    const result = await api.get(
-      '/invitations',
-      isV1Group
-        ? {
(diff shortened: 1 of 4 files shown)


components/group/GroupRelatedInvitations.js
```diff
@@ -11,28 +11,30 @@ const RelatedInvitationRow = ({ item }) => (
...
-      },
-      {
-        prefix: groupId.includes(submissionName) ? `${groupId}/.*` : `${groupId}/-/.*`,
-        expired: true,
-        type: 'all',
-        limit,
-        offset,
-      },
-      { accessToken }
+      isV1Group
```
zbialecki, line 22: I think this would be easier to follow if this wasn't inside the argument list. I would declare a var 
```
const queryParam = isV1Group 
  ? { regex: `${groupId}/-/.*` }
  : { prefix: groupId.includes(submissionName) ? `${groupId}/.*` }
```
and then use that in the call.

zbialecki (approved):

melisabok (approved):

PR #1859, opened 2024-03-15
Feature/ Confirm email page: require user to be logged in

xkopenreview:
this is part of the profile merge update
currently anyone having the email linking url will be able to confirm an email

this pr should change the confirm page to show which email is added and add a captcha
so that when institution user accidently click the email confirm link sent by non-institution account, the institution email won't be auto linked to the non-institution profile.

depends on https://github.com/openreview/openreview-api/pull/385

### DIFF
--- a/pages/confirm.js
+++ b/pages/confirm.js
@@ -6,20 +6,19 @@ import { useRouter } from 'next/router'
 import LoadingSpinner from '../components/LoadingSpinner'
 import ErrorDisplay from '../components/ErrorDisplay'
 import api from '../lib/api-client'
+import useLoginRedirect from '../hooks/useLoginRedirect'
 
 const Confirm = () => {
   const [error, setError] = useState(null)
   const router = useRouter()
+  const { user, accessToken } = useLoginRedirect()
+  const [isLoading, setIsLoading] = useState(true)
+  const [turnstileToken, setTurnstileToken] = useState(null)
+  const [emailToConfirm, setEmailToConfirm] = useState(null)
+  const invalidTurnstileToken = process.env.TURNSTILE_SITEKEY && !turnstileToken
 
-  useEffect(() => {
-    if (!router.isReady) return
-
-    if (!router.query.token) {
-      setError({ statusCode: 404, message: 'Activation token not found' })
-      return
-    }
-
-    api
+  const confirmEmail = async () => {
+    await api
       .put(`/activatelink/${router.query.token}`)
       .then(({ confirmedEmail }) => {
         promptMessage(`Thank you for confirming your email ${confirmedEmail ?? ''}`)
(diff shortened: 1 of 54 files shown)


zbialecki:
I don't think it makes sense to have a captcha here... this page is now going to require the user to log in and then click a button. The captcha was necessary on pages that didn't require authentication, like accepting an invitation, where it's possible a crawler could access the page.

xkopenreview:
another issue related to confirm page which i don't think there's a solution is that when user click the confirm link multiple times, the error message is token is invalid instead of "email is already confirmed"

the activation record is removed from collection when it's confirmed for the first time so the confirm page won't know which email the user is trying to confirm so it can't  be used to check against user's confirmed emails

pages/confirm.js
```diff
@@ -6,20 +6,17 @@ import { useRouter } from 'next/router'
 import LoadingSpinner from '../components/LoadingSpinner'
 import ErrorDisplay from '../components/ErrorDisplay'
 import api from '../lib/api-client'
+import useLoginRedirect from '../hooks/useLoginRedirect'
 
 const Confirm = () => {
   const [error, setError] = useState(null)
   const router = useRouter()
+  const { user, accessToken } = useLoginRedirect()
```
zbialecki, line 14: Hooks are executed in order from top to bottom so `useLoginRedirect` should always be the first line of the function.

pages/confirm.js
```diff
@@ -6,20 +6,17 @@ import { useRouter } from 'next/router'
...
+  const [isLoading, setIsLoading] = useState(true)
```
zbialecki, line 15: So it's considered best practice to not create additional state vars when the value could be derived from other state. Here isLoading could be derived from `emailToConfirm` and `error`.

pages/confirm.js
```diff
@@ -28,11 +25,57 @@ const Confirm = () => {
...
+    <>
+      <p>
+        Click submit button below to confirm adding <strong>{emailToConfirm}</strong> to your
+        account.
+      </p>
+      <p>Please contact info@openreview.net if you didn&apos;t add this email.</p>
 
-  return <LoadingSpinner />
+      <div className="response-buttons">
+        <button type="button" className="btn btn-lg mt-4" onClick={() => confirmEmail()}>
```
zbialecki, line 72:
```suggestion
        <button type="button" className="btn btn-lg mt-4" onClick={confirmEmail}>
```

pages/confirm.js
```diff
@@ -6,20 +6,19 @@ import { useRouter } from 'next/router'
...
-
-    if (!router.query.token) {
-      setError({ statusCode: 404, message: 'Activation token not found' })
-      return
-    }
-
-    api
+  const confirmEmail = async () => {
+    setIsLoading(true)
+    await api
```
zbialecki, line 21: The `await` here isn't necessary if using promises.

pages/confirm.js
```diff
@@ -6,20 +6,19 @@ import { useRouter } from 'next/router'
...
       .put(`/activatelink/${router.query.token}`)
       .then(({ confirmedEmail }) => {
         promptMessage(`Thank you for confirming your email ${confirmedEmail ?? ''}`)
```
zbialecki, line 24: There should be a call to `setIsLoading(false)` here and in the catch.

zbialecki (approved):

zbialecki:
I think this is ready to be merged. It would be nice if the API validated that the user making the request was authenticated, but I don't think it's required.

melisabok:
this font doesn't seem to be matching with the rest of the website, right?

Screen Shot 2024-04-16 at 1 58 46 PM

melisabok:
I think we should add a title: "Account Linking"?

zbialecki:
Yeah, I think a heading on this page would be good. The email should probably be a mailto link as well.

melisabok:
Beside the cosmetic changes, the functionality works as expected. 

I added a test here: https://github.com/openreview/openreview-py/pull/2096

You can run it as:

```
pytest tests/test_profile_management.py::TestProfileManagement::test_confirm_alternate_email
```

melisabok:
What about the button name? should be be "Confirm" instead? or "Confirm email"?

zbialecki:
Good point, I like "Confirm Email"

PR #1861, opened 2024-03-18
Feature/Profile edit: allow the field "ACL Anthology" to be saved in the profile.content.aclanthology

xkopenreview:
this pr should add in profile edit page ACL Anthology field in links section

this pr should also fix failure in setting up test venue v2

### DIFF
--- a/components/profile/PersonalLinksSection.js
+++ b/components/profile/PersonalLinksSection.js
@@ -28,6 +28,15 @@ const PersonalLinkInput = ({ type, links, setLinks }) => {
         setLinks({ type, data: { value, valid: isValid } })
         break
       }
+      case 'aclanthology': {
+        const isValid =
+          isValidURL(value) && value.startsWith('https://aclanthology.org/people/')
+        if (!isValid) {
+          promptError(`${value} is not a valid ACL Anthology URL`, { scrollToTop: false })
+        }
+        setLinks({ type, data: { value, valid: isValid } })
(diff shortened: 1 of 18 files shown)


components/profile/PersonalLinksSection.js
```diff
@@ -28,6 +28,15 @@ const PersonalLinkInput = ({ type, links, setLinks }) => {
         setLinks({ type, data: { value, valid: isValid } })
         break
       }
+      case 'aclanthology': {
+        const isValid =
+          isValidURL(value) && value.startsWith('https://aclanthology.org/people/')
```
zbialecki, line 33: The API has a slightly different rule for validation, the link must also have some text that comes after "people/". Maybe we could just add an additional check `&& value.length > 32`

melisabok:
@xkopenreview can you add an icon with the link to the documentation next to the acl anthology text box? https://docs.openreview.net/getting-started/creating-an-openreview-profile/finding-and-adding-your-acl-anthology-url-to-your-profile

components/profile/PersonalLinksSection.js
xkopenreview: changed to use the same regex as in api so that the general url regex check can be skipped

melisabok (approved):

zbialecki (approved):

haroldrubio (approved):

PR #1863, opened 2024-03-19
Fix/ Note editor: check if invitation has note.content

celestemartinez:
This is needed for this PR: https://github.com/openreview/openreview-py/pull/1998
I am creating an invitation used only to delete/restore submissions. This means the invitation does not have a content

### DIFF
--- a/client/view-v2.js
+++ b/client/view-v2.js
@@ -2305,46 +2305,44 @@ module.exports = (function () {
     })
 
     // content fields
-    if (contentFields) {
-      Object.entries(contentFields).forEach(([contentFieldName, contentFieldValue]) => {
-        if (
-          (formData?.[contentFieldName] === undefined ||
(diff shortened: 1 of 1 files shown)


client/view-v2.js
```diff
@@ -2305,44 +2305,46 @@ module.exports = (function () {
...
-          formData?.[contentFieldName]?.delete === true) &&
-        noteObj?.content?.[contentFieldName]?.value === undefined
-      ) {
-        // do not return field
-        return
-      }
-
-      var valueObj = contentFieldValue.value
-      if (valueObj) {
+    if (contentFields) {
```
melisabok, line 2308: instead of adding a new if, it isn't easier to do `Object.entries(contentFields || [])` ?

melisabok (approved):

xkopenreview (approved):

PR #1865, opened 2024-03-20
Fix/Add profile id to feedback email subject

xkopenreview:
currently all support queries users sent through feedback form about profile id have the same subject which caused them to be grouped to one thread by gmail

this pr should add profile id to subject so that the emails are not grouped


melisabok (approved):

zbialecki (approved):

melisabok:
@xkopenreview can we do that same for "OpenReview Feedback: A conference I organized" and put the venue id?

xkopenreview:
> @xkopenreview can we do that same for "OpenReview Feedback: A conference I organized" and put the venue id?

#1871 should update the subject of a conference i organized and a conference i submitted to

PR #1867, opened 2024-03-20
Fix/ Forum reply - don't show reply invitation button when readers won't match

xkopenreview:
this pr should hide reply button in forum replies when
the invitation note reader has everyone but the parent note is not publicly visible

the readers won't match so that the note editor simply fails to load when the invitation button is clicked

an example is public comment invitation on decision reply or review

### DIFF
--- a/components/forum/ForumReply.js
+++ b/components/forum/ForumReply.js
@@ -11,12 +11,7 @@ import NoteEditor from '../NoteEditor'
 import ForumReplyContext from './ForumReplyContext'
 import Icon from '../Icon'
 import { getInvitationColors } from '../../lib/forum-utils'
-import {
-  prettyId,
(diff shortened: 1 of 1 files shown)


components/forum/ForumReply.js
```diff
@@ -70,6 +65,13 @@ export default function ForumReply({
     }
   }
 
+  const isInvitationVisibleToEveryone = (invitation) => {
+    const invitationReaders = Array.isArray(invitation.edit?.note?.readers)
+      ? invitation.edit?.note?.readers
+      : invitation.edit?.note?.readers?.const
```
melisabok, line 71: do we support `const` inside note.readers?

zbialecki (approved):

melisabok (approved):

PR #1868, opened 2024-03-20
Fix/ Ethics Chair Console: filter out author submissions

celestemartinez:
The Ethics Chair console is still showing the papers submitted by the Ethics Chairs, even though they might not be flagged and shouldn't be visible in the console. This PR filters out the Ethics Chairs' own papers.

### DIFF
--- a/components/webfield/EthicsChairConsole/EthicsChairPaperStatus.js
+++ b/components/webfield/EthicsChairConsole/EthicsChairPaperStatus.js
@@ -80,13 +80,13 @@ const EthicsChairPaperStatus = () => {
         '/notes',
         {
           invitation: submissionId,
-          details: 'replies',
+          details: 'replies,writable',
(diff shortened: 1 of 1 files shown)


melisabok (approved):

zbialecki (approved):
Are Ethics Chairs always different from Program Chairs? Because with this change if a PC tries to view this console they won't see any notes.

xkopenreview:
> Are Ethics Chairs always different from Program Chairs? Because with this change if a PC tries to view this console they won't see any notes.

maybe it should check author instead of writable?
i think right now ethics chairs of existing conferences are not pc but not sure about future conferences

melisabok:
> maybe it should check author instead of writable?

That wouldn't work for single blind venues.

melisabok:
If the Ethics Chairs are PCs then they should use PC console to see the papers.

xkopenreview (approved):

PR #1869, opened 2024-03-20
Fix/DBLP import - Allow importing from venues that is hosted by openreview

xkopenreview:
currently dblp import of papers from venues hosted on openreview is not allowed if the venue in dblp and in openreview is the same.

this pr should change it to only check for previously imported dblp papers instead of all papers associated with authors (those submitted to a venue) so that the dblp paper can be imported

this pr should also add referrer param to papers imported from dblp so that the banner can navigate back to user's profile edit page which currently navigates to non-existing group

### DIFF
--- a/lib/profiles.js
+++ b/lib/profiles.js
@@ -413,7 +413,7 @@ export async function getAllPapersByGroupId(profileId, accessToken) {
   try {
     const notesList = await api.get(
       '/notes',
-      { 'content.authorids': profileId },
+      { 'content.authorids': profileId, invitation: 'dblp.org/-/record' },
       { accessToken, version: 1 }
     )
     return notesList.notes.map((note) => ({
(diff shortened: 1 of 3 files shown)


lib/profiles.js
```diff
@@ -413,7 +413,7 @@ export async function getAllPapersByGroupId(profileId, accessToken) {
   try {
     const notesList = await api.get(
       '/notes',
-      { 'content.authorids': profileId },
+      { 'content.authorids': profileId, invitation: 'dblp.org/-/record' },
```
melisabok, line 416: we will user another invitation for API 2 import, so we will need to change this later.

xkopenreview: yes it will be changed in the dblp import using v2 pr after this one is merged

xkopenreview: updated in https://github.com/openreview/openreview-web/pull/1716/commits/ab87d40cc59eaf86d0b91781a5e1a94ebd28756e

melisabok (approved):

zbialecki (approved):

racheljsmart (approved):

PR #1876, opened 2024-04-01
SAC Console: hide AC tab when there is no assignment invitation

celestemartinez:
We now allow SACs to be directly assigned to papers instead of to ACs. When the webfield does not contain the `assignmentInvitation`, then we should hide the 'Area Chair Status' tab.

### DIFF
--- a/components/webfield/SeniorAreaChairConsole.js
+++ b/components/webfield/SeniorAreaChairConsole.js
@@ -643,6 +643,7 @@ const SeniorAreaChairConsole = ({ appContext }) => {
             id="areachair-status"
             active={activeTabId === '#areachair-status' ? true : undefined}
             onClick={() => setActiveTabId('#areachair-status')}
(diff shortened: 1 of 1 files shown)


melisabok:
There are some ES lint errors. You can run `npm run lint` to see them.

melisabok (approved):

components/webfield/SeniorAreaChairConsole.js
```diff
@@ -643,6 +643,7 @@ const SeniorAreaChairConsole = ({ appContext }) => {
             id="areachair-status"
             active={activeTabId === '#areachair-status' ? true : undefined}
             onClick={() => setActiveTabId('#areachair-status')}
+            hidden={assignmentInvitation ? false : true}
```
zbialecki, line 646:
```suggestion
            hidden={!assignmentInvitation}
```

PR #1878, opened 2024-04-02
Fix/Ethics Chair console - anonymous group data loading logic update

xkopenreview:
currently ethics chair console expect the per paper ethics reviewers group to contain the tilde ids of ethics reviewers
while the new logic has changed to member of per paper ethics reviewers group to the anonymous id group of each assigned ethics reviewer

this pr should update the mapping logic accordingly
the changes should be compatible with the case where tilde ids are added as member of ethics reviewers group directly (for example NeurIPS 2023)

this pr should also fix query searching ethics reviewers by name or email (right now this is not working properly)


celestemartinez (approved):
Looks good! I can see the assigned ethics reviewers.

zbialecki (approved):

PR #1882, opened 2024-04-02
Fix/ Profile page: show more information to help identify fake profiles

### LINKED ISSUE #1879: Profile page: show when the profile was created
As a security measure to avoid fake profiles to be created in OpenReview. The policy board committee suggested to show the profile created date in the public profile page.

I suggest to show is next to the name as "created on ..." "user since....", "joined ..."

### LINKED ISSUE #1880: Profile Page: show confirmed emails in the public view
Currently only the super user and profile owner see the "confirmed" label.

PCs would like to see which institutional emails are confirmed in order to help detect fake profiles.

### LINKED ISSUE #1881: Profile Page: relations not correctly rendered when switching profiles
1. Go to the profile: https://openreview.net/profile?id=~Margaret_M._Fleck2
2. Click on relation "[David Forsyth](https://openreview.net/profile?id=~David_Forsyth1)"
3. Click on the back button

Margaret's profile is showing relations from David's profile and I believe the issue is because these relations are duplicates.

zbialecki:
- Show which emails are confirmed to all users. Fixes #1880 
- Show the date when the profile was created. Fixes #1879 
- Update relations properly when switching profiles. Fixes #1881

### DIFF
--- a/components/profile/BasicProfileView.js
+++ b/components/profile/BasicProfileView.js
@@ -222,8 +222,7 @@ const BasicProfileView = ({
               <ProfileRelation
                 key={
                   relation.relation +
-                  relation.name +
-                  relation.email +
+                  relation.username +
                   relation.start +
                   relation.end
                 }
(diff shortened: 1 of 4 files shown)


melisabok (approved):
looks good to me

zbialecki:
Thanks, I think this should be fixed now since all relations should have a name even if missing a username.

PR #1884, opened 2024-04-03
Fix/ Tasks page: mark task as incomplete if the reply is deleted

melisabok:
It seems that check to mark as task as complete is the existence of the `repliedEdits` but some of these edits could be deleting the reply note. 

If a review is deleted, the task still appears a complete and it should be back to pending.

This PR checks the invitation has repliedEdits and the last edit doesn't content a note.ddate assuming the edits are sorted chronologically. @carlosmondra please confirm this is the case.

I created a new test case for API 2 tasks.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,7 +18,6 @@ import get from 'lodash/get'
 import intersection from 'lodash/intersection'
 import union from 'lodash/union'
 import uniq from 'lodash/uniq'
-import { min } from 'lodash'
 
 /**
  * Get the human readable version of a group or profile id
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -777,7 +777,7 @@ export function formatTasksData(
     if (inv.noteInvitation) {
       if (
         inv.details.repliedNotes?.length ||
-        (inv.apiVersion === 2 && inv.details.repliedEdits?.length)
+        (inv.apiVersion === 2 && inv.details.repliedEdits?.length && !inv.details.repliedEdits.slice(-1)[0].note.ddate)
       ) {
         inv.completed = true
       }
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -777,7 +777,7 @@ export function formatTasksData(
     if (inv.noteInvitation) {
       if (
         inv.details.repliedNotes?.length ||
-        (inv.apiVersion === 2 && inv.details.repliedEdits?.length)
+        (inv.apiVersion === 2 && inv.details.repliedEdits?.length && !inv.details.repliedEdits.slice(-1)[0].note.ddate)
```
zbialecki, line 780: What if there was an invitation with `minReplies` > 1? Should we be counting note replies here?

melisabok: minReplies is used for notes not for edits. 

I don't think we are checking that in the code....

zbialecki: Couldn't we iterate over the `repliedEdits` and get a count of the notes?

melisabok: I'm not sure about this approach because there can be multiple edits for the same note and we would need to check if the latest edit doesn't delete the note. 

AFAIK we are not using minReplies > 1 for note invitations so I would create an issue and fix this in another PR.

zbialecki (approved):

carlosmondra:
> This PR checks the invitation has repliedEdits and the last edit doesn't content a note.ddate assuming the edits are sorted chronologically. @carlosmondra please confirm this is the case.

This is not guaranteed. I would recommend sorting them in the UI.

zbialecki (approved):

zbialecki:
@xkopenreview Can you take a look? You can test by replying to any note invitation and then deleting that reply.

PR #1891, opened 2024-04-05
Fix/ Add turnstile token to password reset

xkopenreview:
This pr should add turnstile widget to password reset page
depends on https://github.com/openreview/openreview-api/pull/401

this pr should also extract turnstile widget related logic to a hook so that it can be added to other pages

### DIFF
--- a/tests/registerPage.ts
+++ b/tests/registerPage.ts
@@ -589,11 +589,7 @@ test('reset password of active profile', async (t) => {
   await t
     .typeText(Selector('#email-input'), 'melisa@test.com')
     .expect(Selector('button').withText('Reset Password').hasAttribute('disabled')).notOk({ timeout: 5000 })
-
-  await t
-    .wait(5000)
     .click(Selector('button').withText('Reset Password'))
-    .wait(5000)
     .expect(Selector('div').withAttribute('role', 'alert').exists)
     .ok()
 


hooks/useTurnstileToken.js
```diff
@@ -0,0 +1,33 @@
...
+  useEffect(() => {
+    if (!renderWidget) {
+      setTurnstileToken(null)
+      if (widgetId) {
+        window.turnstile.remove(widgetId)
+      }
+      return
+    }
+    if (window.turnstile) {
+      const id = window.turnstile.render(`#turnstile-${key}`, {
```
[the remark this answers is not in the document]
zbialecki: Also, you could use the state of the ref to replace the renderWidget prop. If ref is null, that means the container element is not being rendered by the parent.

components/FeedbackModal.js
```diff
@@ -7,20 +7,21 @@ import ErrorAlert from './ErrorAlert'
...
-  const [turnstileToken, setTurnstileToken] = useState(null)
   const [error, setError] = useState(null)
+  const [isOpen, setIsOpen] = useState(false)
   const { accessToken } = useContext(UserContext)
   const [formData, setFormData] = useReducer((state, action) => {
     if (action.type === 'reset') return {}
     if (action.type === 'prefill') return action.payload
     return { ...state, [action.type]: action.payload }
   }, {})
+  const { turnstileToken, turnstileContainerRef } = useTurnstileToken('feedback')
```
zbialecki, line 23: I was actually suggesting the parent component manage the ref and pass it to the hook. Then the parent component could conditionally render the container using some state var. Would this pattern make the code simpler?

pages/reset.js
```diff
@@ -50,8 +52,12 @@ const ResetForm = ({ setEmailSent }) => {
           onChange={(e) => setEmail(e.target.value.trim())}
         />
       </div>
-
-      <button type="submit" className="btn btn-primary" disabled={!isValidEmail(email)}>
+      <div ref={turnstileContainerRef} />
```
zbialecki, line 55: Should be an equal margin below this div as there is below the email input

zbialecki (approved):
Looks like it's working as expected, just a small comment

melisabok (approved):

carlosmondra (approved):

PR #1894, opened 2024-04-09
Feature/ User moderation - allow moderator to search and accept a profile

xkopenreview:
currently accept profile action button is only available in the moderate only queue and not available in the all profiles queue
so when there are many profiles pending moderation it's tedious for the moderator to find a profile and activate it.

this pr should allow moderator to search a pending moderation profile from all profiles and accept it.
delete button is hidden for needs moderation profiles to make space for accept button

this pr should also sort the profiles pending moderation by modification date so that those old profiles which are claimed can be seen

this pr should also color the modification date of the profiles which has passed 2 weeks

### DIFF
--- a/pages/user/moderation.js
+++ b/pages/user/moderation.js
@@ -1756,11 +1756,13 @@ const UserModerationQueue = ({
     'Inactive',
     'Merged',
   ].map((p) => ({ label: p, value: p }))
+  const twoWeeksAgo = dayjs().subtract(2, 'week').valueOf()
 
   const getProfiles = async () => {
     const queryOptions = onlyModeration ? { needsModeration: true } : {}
     const cleanSearchTerm = filters.term?.trim()
     const shouldSearchProfile = profileStateOption === 'All' && cleanSearchTerm
+    const sortKey = onlyModeration ? 'tmdate' : 'tcdate'
     let searchQuery = { fullname: cleanSearchTerm?.toLowerCase() }
     if (cleanSearchTerm?.startsWith('~')) searchQuery = { id: cleanSearchTerm }
     if (cleanSearchTerm?.includes('@')) searchQuery = { email: cleanSearchTerm.toLowerCase() }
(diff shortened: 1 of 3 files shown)


melisabok (approved):

melisabok:
@zbialecki @racheljsmart please review this and merge it, thanks!

pages/user/moderation.js
```diff
@@ -2028,7 +2030,35 @@ const UserModerationQueue = ({
...
+                      month: 'short',
+                      year: '2-digit',
+                      hour: '2-digit',
+                      minute: '2-digit',
+                      second: undefined,
+                      timeZoneName: undefined,
+                      hour12: false,
+                    })}
+                  </span>
+                  /
```
zbialecki, line 2048: There should be spaces around this separator.

pages/user/moderation.js
```diff
@@ -2028,7 +2030,35 @@ const UserModerationQueue = ({
...
+                  <span>
+                    {formatDateTime(profile.tcdate, {
```
zbialecki, line 2050: What about only displaying 1 date if the dates are the same? It would make it much more obvious which profiles have different created/modified dates.

pages/user/moderation.js
```diff
@@ -2028,7 +2030,35 @@ const UserModerationQueue = ({
                   </a>
                 </span>
                 <span className="col-email text-muted">{profile.content.preferredEmail}</span>
-                <span className="col-created">{formatDateTime(profile.tcdate)}</span>
+                <span className="col-created">
+                  <span
+                    className={`${profile.tmdate < twoWeeksAgo ? 'passed-moderation' : ''}`}
+                  >
+                    {formatDateTime(profile.tmdate, {
```
zbialecki, line 2037: I think it would make more sense for the created date to go first - it will always be the first date chronologically.

zbialecki (approved):
Looks good to me

melisabok:
merging this now and asking Rachel to test this in the dev site.

PR #1896, opened 2024-04-09
Fix/ Group directory: sort child groups correctly

### LINKED ISSUE #1895: Group directory page: sort child groups alphabetically
In this case, there are venues with the same prefix that are not shown next to each other:

https://openreview.net/group?id=aclweb.org/ACL/2024/Workshop

zbialecki:
Fix alphabetical sort. Add case for groups with matching years.

Fixes #1895


melisabok (approved):
Thanks!

PR #1897, opened 2024-04-12
Feature/ PC Console: Add paper status tabs filtered by content field

haroldrubio:
This PR allows flagged submissions to be shown in a separate tab, along with the replies that raised the flag and replies that were made in response to the flag. Some of the use cases are:
- (ARR) Reviewers and ACs run through a quick checklist of their assigned papers before reviewing begins to identify desk rejections. The PCs need a way to check if the reason the flags were raised are valid, and if it was already looked at by another PC.
- Papers can get flagged for ethics reviews by an official review and the reviewer can explain the reason for the flag. The PCs can check if an ethics review has already been submitted to these papers.
- Papers can get nominated and flagged for an award by ACs/SACs. The PCs can award a paper with another forum reply, where it can be tracked in this new tab.

This PR makes the following changes:

- Adds a new field in the webfieldConfig `submissionContentFields` which is a list of objects, one per tab, that describes the data shown
- Creates a new tab based on each object in `submissionContentFields`, modifies the menu bar to sort by the `field` attribute and exports that field in the CSV
- Replaces the decision column with the extra information surrounding the flagging

Example configuration for ICML ethics reviewing:
```
submissionContentFields: [
  {
    field: 'flagged_for_ethics_review',
    responseInvitations: ['Ethics_Review'],
    reasonInvitations: ['Ethics_Review_Triage'],
    reasonFields: {
      'ethics_review_triage': ['Ethics review needed.']
    }
  }
]
```
- `field`: The submission content field that the tab will be based on. The tab will filter based on whether or not this field is present in a submission's content
- `responseInvitations`: Array of invitation endings that are used to display replies that were made after, and in response to the submission being flagged
- `reasonInvitations`:  Array of invitation endings that are used to display replies that were made before, or in parallel, the flag and are the reason that the flag was raised
- `reasonFields`: Object that contains content fields that may be in any of the notes that reply to any of the reason invitations. These are the fields and possible values that would cause the process function to flag the submission


zbialecki:
Seems like you have a lint error in `./components/webfield/ProgramChairConsole/PaperStatus.js`

melisabok (approved):

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -143,13 +276,25 @@ const PaperStatus = ({ pcConsoleData, loadReviewMetaReviewData }) => {
     } else {
       const { notes, noteNumberReviewMetaReviewMap } = pcConsoleData
       if (!notes) return
-      const tableRows = [...(noteNumberReviewMetaReviewMap.values() ?? [])]
+      const actualNotes = noteContentField
+        ? notes.filter((note) => Object.keys(note.content).includes(noteContentField.field))
```
zbialecki, line 280: In JavaScript to check if an object has a certain field you can just do `note.content[noteContentField.field]`. Since note content fields are always objects, if this returns false you know the note doesn't have that field.

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -143,13 +276,25 @@ const PaperStatus = ({ pcConsoleData, loadReviewMetaReviewData }) => {
...
+        : notes
+      const actualNoteNumbers = actualNotes.map((note) => note.number)
+      const actualNoteNumberReviewMetaReviewMap = noteContentField
+        ? new Map(
+            [...noteNumberReviewMetaReviewMap].filter(([noteNumber, dataMap]) =>
```
zbialecki, line 285: filter does not modify the original array so you don't need to make a copy, you can just do noteNumberReviewMetaReviewMap.filter

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -143,13 +276,25 @@ const PaperStatus = ({ pcConsoleData, loadReviewMetaReviewData }) => {
...
+              actualNoteNumbers.includes(noteNumber)
+            )
+          )
+        : noteNumberReviewMetaReviewMap
+
+      const tableRows = [...(actualNoteNumberReviewMetaReviewMap.values() ?? [])]
```
zbialecki, line 291: Same here... can just do
```suggestion
      const tableRows = actualNoteNumberReviewMetaReviewMap.values()
```

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -99,15 +131,116 @@ const PaperRow = ({
...
+                        note={reasonNote}
+                        referrerUrl={referrerUrl}
+                        showReaders={false}
+                        isV2Note={true}
+                      />
+                    </td>
+                  </tr>
+                ))}
+              </Table>
+              <hr></hr>
```
zbialecki, line 188: In React you just use self-closing tags: `<hr />`

components/webfield/ProgramChairConsole/PaperStatus.js
zbialecki, line 291: `values` will always return an iterator, so you don't need the fallback, also I would probably use Array.from here to make it more clear you're converting a Map iterator.
```suggestion
      const tableRows = Array.from(actualNoteNumberReviewMetaReviewMap.values())
```

melisabok:
@xkopenreview @zbialecki you can test this change using https://github.com/openreview/openreview-py/pull/2121

you need to run `pytest tests/test_arr_venue_v2.py`

zbialecki (approved):
Looks good to me

melisabok:
I got this error trying to export the paper status tab:

Screen Shot 2024-05-02 at 3 50 13 PM

melisabok:
When I click on export in the flag tab, the name of the file is "....Paper Status.csv", may be we should put the name of the tab instead.

haroldrubio:
The export error I think is the same error as the sort one, I need to apply the fix to that element in the export array

melisabok:
Looks good to me.

melisabok:
@haroldrubio do you want to address Harold's comment or should we merge it?

PR #1898, opened 2024-04-12
Fix/ DBLP import - allow up to 500 papers to be imported each time

xkopenreview:
when a lot of papers (>1000) are selected for import
the browser will report error for insufficient resource
api will report error for max reply reached
cloud function to abstract extraction may throw service unavailable error (reason is unknown)

so this pr should allow user to select up to 500 papers for import to avoid error so that user don't keep retry importing which  may cause the process function queue to be blocked


melisabok (approved):

melisabok:
@zbialecki do you want to take a look?

zbialecki (approved):

PR #1904, opened 2024-04-19
Fix/ Forum page: add pagination for forums with many replies 

### LINKED ISSUE #1875: Forum page - user is prompted to login when there's no action
https://github.com/openreview/openreview-web/assets/60613434/ecc93fb0-ba2c-4565-a433-10c80a1b3d24

### LINKED ISSUE #1892: Forum page: add a tooltip to all the readers selection options
They all start with the same prefix and it is hard to know which group id corresponds to each reviewer id:

Screen Shot 2024-04-08 at 9 20 46 AM

zbialecki:
Also adds tooltips to the readers filter and removes the login prompt when unauthenticated users load the page. Users will only be prompted to login when they click an invitation button.

Fixes #1892 
Fixes #1875

### DIFF
--- a/components/NoteEditor.js
+++ b/components/NoteEditor.js
@@ -562,14 +562,12 @@ const NoteEditor = ({
   }
 
   useEffect(() => {
-    if (userLoading) return
+    if (userLoading || !invitation?.edit?.note?.content) return
 
     if (!user) {
       promptLogin()
       return
     }
-
-    if (!invitation?.edit?.note?.content) return
 
     setFields(
       Object.entries(invitation.edit.note.content).sort(
--- a/components/form/ToggleButtonGroup.js
+++ b/components/form/ToggleButtonGroup.js
@@ -1,3 +1,4 @@
+/* globals $: false */
 import { useEffect, useState } from 'react'
 import truncate from 'lodash/truncate'
 import findIndex from 'lodash/findIndex'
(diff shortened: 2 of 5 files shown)


melisabok:
The changes look good, it seems to be fixing both issues. 

I tested the pagination with the TMLR request forum that has more than 2k replies: http://localhost:3030/forum?id=xLdUNpcuxA

When I filter by "Visible To", the replies are filtered correctly but when I try to remove the filter it gets stuck and it takes a lot of seconds to show all the replies again. 

You can reproduce it by going to the TMLR forum, click on visible to "everyone" and then remove the token "everyone".

zbialecki:
I'll look into it. You're only seeing that issue on TMLR request forum page right? Other regular forum pages are working normally?

melisabok:
NeurIPS Reviewer registration also fails: http://localhost:3030/forum?id=DUG81c4rFc that has more than 7k replies. The reply count seems to be off showing -5. 

Screen Shot 2024-04-19 at 11 33 46 AM

zbialecki:
I tried out a number of different performance optimizations that had some effect but I'm still noticing lagging especially on the CVPR forum with 7000 replies. For now I think this is a pretty extreme edge case so it should be OK.

@xkopenreview I'm not sure what you mean by "i think it will cause cutoff index to be 0 for the first few clicks of view more replies". For me clicking view more shows an additional 50 replies as expected.

zbialecki:
@xkopenreview Can you test your example forum again? It should be fixed now.

melisabok (approved):

PR #1905, opened 2024-04-19
Fix/Moderation page - update reject reason and show messages

xkopenreview:
this pr should solve the following 2 issues:
1. the reject reason does not fit actual use case
   invalid and inpersonal homepage options are combined
   added options for institution email validation and for keep submitting invalid info
2. it's hard to know whether the profile has been rejected by other moderators before
   added the 5 most recent messages in profile preview so that moderator does not need to go to messages page and paste the email to see the messages sent

### DIFF
--- a/components/ExpertiseSelector.js
+++ b/components/ExpertiseSelector.js
@@ -138,6 +138,27 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
     )
   }
 
+  const markTaskAsComplete = async () => {
+    try {
+      await api.post(
+        '/edges',
+        {
+          invitation: invitation.id,
+          readers: [venueId, user.profile.id],
+          writers: [venueId, user.profile.id],
+          signatures: [user.profile.id],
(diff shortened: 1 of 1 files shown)


melisabok (approved):

components/profile/MessagesSection.js
```diff
@@ -0,0 +1,72 @@
...
+import { useEffect, useState } from 'react'
+import api from '../../lib/api-client'
+import Table from '../Table'
+
+const MessagesSection = ({ email, accessToken }) => {
+  const [messages, setMessages] = useState([])
+  const [isLoading, setIsLoading] = useState(true)
+
+  const loadMessages = async () => {
+    try {
```
zbialecki, line 11: It would be good to setLoading to true at the start of the function, so if the email changes the UI shows the loading indicator.

zbialecki (approved):

enrubio (approved):
I think this looks good and I like the updated templates.

Screenshot 2024-04-23 at 3 20 09 PM

PR #1915, opened 2024-04-24
Fix/Edge browser - fix entity with 0 quota

xkopenreview:
currently when the custom quota is 0, the quotaNotReached value will be wrong
and it will cause the entity to appear in column even if the filter quota reached checkbox is checked

this pr should fix this issue

### DIFF
--- a/components/browser/Column.js
+++ b/components/browser/Column.js
@@ -589,7 +589,7 @@ export default function Column(props) {
       const { filteredRows, queryIsInvalid } = filterCollections(
         colItems.map((p) => {
           const customLoad = getQuota(p)
-          const quotaNotReached = !customLoad || p.traverseEdgesCount < customLoad
+          const quotaNotReached = customLoad === undefined || p.traverseEdgesCount < customLoad
 
           return {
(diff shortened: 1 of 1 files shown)


zbialecki (approved):

components/browser/Column.js
```diff
@@ -589,7 +589,7 @@ export default function Column(props) {
       const { filteredRows, queryIsInvalid } = filterCollections(
         colItems.map((p) => {
           const customLoad = getQuota(p)
-          const quotaNotReached = !customLoad || p.traverseEdgesCount < customLoad
+          const quotaNotReached = customLoad === undefined || p.traverseEdgesCount < customLoad
```
melisabok, line 592: If there is no custom load I think we don’t have to check this so I would remove the first part of the or

melisabok: I know, there are some reviewers that appear in the column that are not members of the reviewers group so they don't have a custom max papers edge so I think we shouldn't show them as available.

melisabok (approved):

PR #1918, opened 2024-04-29
Feature/update token issuer

carlosmondra:
This PR checks that the issuer of the token starts with `openreview`. The issuer value will change depending on the timestamp of the terms and conditions.


zbialecki (approved):
Looks good

melisabok (approved):

PR #1934, opened 2024-05-06
Fix/PC Console - Allow functions for search query

xkopenreview:
this pr should add support for function strings so that the query search can search for arbitrary field

### DIFF
--- a/components/webfield/ProgramChairConsole/PaperStatusMenuBar.js
+++ b/components/webfield/ProgramChairConsole/PaperStatusMenuBar.js
@@ -12,7 +12,7 @@ const PaperStatusMenuBar = ({
   selectedNoteIds,
   setPaperStatusTabData,
   reviewRatingName,
-  noteContentField
+  noteContentField,
 }) => {
   const {
     apiVersion,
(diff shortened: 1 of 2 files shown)


components/webfield/ProgramChairConsole/PaperStatusMenuBar.js
```diff
@@ -72,9 +72,46 @@ const PaperStatusMenuBar = ({
         }),
         {}
       )),
-    ...(typeof extraPropertiesAllowed === 'object' && extraPropertiesAllowed),
+    ...(typeof extraPropertiesAllowed === 'object' &&
+      Object.fromEntries(
+        Object.entries(extraPropertiesAllowed).map(([key, value]) => {
+          if (typeof value === 'string') {
+            return [key, [key]]
```
zbialecki, line 79: Can you explain what this is doing? Why is it discarding value if it's a string?

zbialecki (commented):
My issue with this PR is that this pattern you're using of `Object.fromEntries(Object.entries(...).map(...))` isn't efficient (because of the multiple iterations and allocations) and it's harder to read than having a simple forEach or using _.mapValues. It's especially important to optimize this when you're iterating over all rows, like you are on line 104.

melisabok:
@xkopenreview can you address the comments here so we can use this PR for the next ARR June cycle?

melisabok:
ohhh I see the commit. 

Could you put an example of the configuration so I can test it using a test in openreview-py?

melisabok:
@haroldrubio is it possible this must be implemented in the SAC console too? or only the PC console needs it?

melisabok (approved):

PR #1938, opened 2024-05-08
Fix/ PC Console: Get registration notes for all committee members

haroldrubio:
Fixes an issues where the registration notes were only being fetched for assigned members and not all group members.

Emails may need to be sent out before any assignments are made because the registration notes can contain information that is used for the matching


haroldrubio:
This change will be pretty expensive for large venues since it ends up searching for a lot more profiles. I'm not sure if this is worth it - maybe it would be better to just get the profiles for all signatures of the registration notes?

My concern with getting profiles based on the signatures instead is that ARR may end up removing some reviewers after they've made a registration note so: there will be signatures that are no longer members of the reviewer group

melisabok:
Fixed in another PR

PR #1942, opened 2024-05-09
Feature/PC Console - Enable SAC Emailing

haroldrubio:
This PR enables emailing from the SAC status tab in the PC console. Also supports a new field in the webfield config to add custom filtering functions: `sacEmailFuncs`. The changes are mainly copying the messaging blocks of code from the AreaChairStatusMenuBar

Example:
```
sacEmailFuncs: [
      {
        label: 'Assigned Senior Area Chairs', filterFunc: `
        return row.acs.length > 0
        `
      }
    ]
```

May need to depend on some changes in `openreview-py` to support this field in the webfield config:
`messageAreaChairsInvitationId: domain.content.area_chairs_message_id?.value,`


melisabok:
@haroldrubio can you address the comments from Xukun so we can merge this PR? I would like to add this config to the ARR webfield in openreview-py. Maybe you can create a PR with the config to test and then we can merge this.

melisabok:
Fixed in another PR

PR #1943, opened 2024-05-09
Fix/PC Console - Fixes SAC Status Tab for SAC Paper Assignments

haroldrubio:
This PR fixes an issue where the SAC status tab does not display correct information. This PR just restores the information as if there were SAC and AC assignments, it does not show information about the submissions assigned to the SAC.

This PR changes how the edges are parsed, and adds an API call to get the AC assignment edges if the variable `enableSacPaperAssignments: true` is set.


melisabok:
No needed, it was fixed in another PR.

PR #1945, opened 2024-05-09
Fix/tests in openreview-py about custom quotas

melisabok:


melisabok:
> should it be a number when saving this note?

Yes, we need to use numbers instead of string to store the reduced_load value, I will create an issue.

PR #1949, opened 2024-05-11
feat: expertise selection ui

RonLek:
Feat:
- add colored expertise keyphrases.
- show authored papers bucketed by clicked keyphrase with paper weighted heatmap.
- dynamic potential paper updates on keyphrase include/exclude and add/delete.


melisabok:
It seems you are overwriting the default expertise implementation. I suggest to implement a new component for your implementation.

zbialecki:
I agree with Melisa here. When we discussed this last I thought I made clear that you needed to create a new component with a different (descriptive) name that would be enabled by changing the webfield configuration for the venue group.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
+      setActivePapers(jsonResponse['profilekp2user_papers'])
+      // in the future, paper weights should be set based on id instead of titles
+      setPaperWeights(buildPaperWeightDict(jsonResponse))
+    }
+    fetchRecommendations()
+  }, [includedKeyphrases])
+
+  useEffect(() => {
+    keyphrases &&
+      setKeyphraseColors(
```
zbialecki, line 103: These colors could be computed when the keyphrases are loaded and stored in the same data structure. In general you want to minimize the number of depended state vars and useEffects.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
-    <Tabs className={styles.container}>
-      <TabList>
-        <Tab id="all-your-papers" icon="search" active>
-          All Your Papers
-        </Tab>
-        <Tab id={tabId} headingCount={selectedIds?.length}>
-          {tabLabel}
-        </Tab>
-      </TabList>
+    <>
```
zbialecki, line 144: You don't need the fragment, the div is the container for the whole component.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
+                    className="col-sm-2"
+                    name="floppy-save"
+                    text="Save Expertise"
+                    onClick={() => alert('Expertise saved!')}
+                  />
+                </li>
+              </ol>
+            </div>
+            <div>
+              {keyphrases['active'].map((keyphrase, index) => {
```
zbialecki, line 172: You should run `npm run lint` to catch these formatting errors

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
+            </div>
+          </div>
+
+          <div className="col-md-9 paper-container">
+            <div className="row column-header">
+              <h4 className="col-sm-10">Potential Paper Assignments</h4>
+            </div>
+            <div className="mb-3 ml-3">
+              Potential assignments from papers accepted in machine learning conferences from
+              2018 to 2023
```
zbialecki, line 291: This text and these dates shouldn't be hardcoded, it could come from config vars or another source

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
+            </div>
+
+            {recommendedPapers.map((paper) => {
+              return (
+                <>
```
zbialecki, line 296: Unnecessary fragment

components/Icon.js
```diff
@@ -1,10 +1,11 @@
-const Icon = ({ name, tooltip, extraClasses = '' }) => (
+const Icon = ({ name, tooltip, extraClasses = '', onClick}) => (
```
zbialecki, line 1: This component should not be modified - to add a click handler you should create a button that contains the icon. The icon itself should not be clickable

styles/components/Icon.module.scss
```diff
@@ -0,0 +1,11 @@
+.blue{
```
zbialecki, line 1: These styles should go in your components stylesheet. See constants.scss for app colors.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
-          '/notes',
-          {
-            'content.authorids': user.profile.id,
-            sort: 'cdate',
-            details: 'invitation',
-          },
-          null,
-          { accessToken, includeVersion: true }
+        const keyphraseResponse = await fetch(
+          'https://retrievalapp-zso5o2q47q-uc.a.run.app/get_recommendations/mccallum'
```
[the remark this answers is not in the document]
RonLek: Our backend model currently only has the data for a few limited users and might not support all users with an openreview account at present. @MSheshera might be able to share more information.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
-        const { notes } = await api.getCombined(
-          '/notes',
-          {
-            'content.authorids': user.profile.id,
-            sort: 'cdate',
-            details: 'invitation',
-          },
-          null,
-          { accessToken, includeVersion: true }
+        const keyphraseResponse = await fetch(
```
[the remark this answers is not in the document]
RonLek: The current implementation of the `api` lib seems to be for relative requests on the openreview-api. It prefixes the `path` with a `/` if it doesn't begin with it already. Additionally, the `baseUrl` used is the `API_URL`. I'm looking at the implementation [here](https://github.com/openreview/openreview-web/blob/94bd7dcb4a1a129c56b4410c88bc55d8ec07ceb7/lib/api-client.js#L63-L67). Please let me know if I should modify the lib to allow making absolute requests to other domains.

components/ExpertiseSelector.js
```diff
@@ -160,112 +35,330 @@ export default function ExpertiseSelector({ invitation, venueId, apiVersion, sho
...
+    keyphrases &&
+      setKeyphraseColors(
+        keyphrases['active'].reduce((acc, kp) => {
+          acc[kp] = stringToColor(kp)
+          return acc
+        }, {})
+      )
+  }, [keyphrases])
+
+  function hashString(s) {
```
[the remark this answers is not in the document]
RonLek: Thanks! Moved into `lib/utils.js`. [lib/forum-utils](https://github.com/openreview/openreview-web/blob/94bd7dcb4a1a129c56b4410c88bc55d8ec07ceb7/lib/forum-utils.js#L274-L307) has a `getInvitationColors` which I took inspiration from but it seems limited to 15 bg colors. For the controllable expertise selector UI we currently limit to upto 20 keyphrases (each requiring a unique color). Let me know your thoughts!

components/ExpertiseSelector.js
RonLek: Thanks for the suggestion! Moved to the on load useEffect.

components/Icon.js
RonLek: Thanks! Made use of `IconButton` with adjusted padding instead. Slightly changes from the previous UI (with borders and colors) but I think it fits the theme more closely.

styles/components/Icon.module.scss
RonLek: Removed this file in the latest revision.

components/ExpertiseSelector.js
RonLek: Thanks! This `div` might get removed in a future iteration. If not where do you suggest I get this? I couldn't find a specific `config` var in the `openreview-web` codebase, however I'm aware of `config` within the v1/v2 APIs being present.

components/webfield/ExpertiseConsole.js
```diff
@@ -21,6 +22,7 @@ export default function ExpertiseConsole({ appContext }) {
   } = useContext(WebFieldContext)
   const [shouldReload, reload] = useReducer((p) => !p, true)
   const router = useRouter()
+  const useControllableExpertiseSelector = true
```
melisabok, line 25: this value must be a parameter of the ExpertiseConsole and it should be false by default.

components/ControllableExpertiseSelector.js
```diff
@@ -0,0 +1,299 @@
...
+            }
+          }
+        })
+      })
+
+      return paperWeightDict
+    }
+    async function fetchRecommendations() {
+      const baseUrl =
+        'https://retrievalapp-zso5o2q47q-uc.a.run.app/get_recommendations/mccallum'
```
melisabok, line 74: can this be a parameter of the component?

does it work for all the venues or only specific ones?

MSheshera: Actually this will not work for all venues. This backend caches reviewer data and reviewers will be specific to a venue. Venues may also choose to show "Potential Paper Assignments" using a domain-specific collection e.g. while NLP/ML venues may use papers from those veneues (as we are now) Computer Vision venues may use a different collection of papers.

This can be changed in future designs but for a first run, this is easiest to set up.

melisabok: How do you handle authentication here? is this link public for any user?

components/webfield/ExpertiseConsole.js
melisabok: This change is missing, could you please add it so we can merge this soon?

components/ControllableExpertiseSelector.js
melisabok: @carlosmondra indicates that this server must support cross domain so it can be called from openreview.net.

MSheshera: > How do you handle authentication here? is this link public for any user?

We havent added any authentication to the API call yet. This is one of our immediate todos - would you prefer for us to complete this before this PR is accepted?

> this server must support cross domain so it can be called from openreview.net.

What do you mean? It should be possible to call the server from openreview.net.

MSheshera:
> @MSheshera I don't find the API call that stores the keyphrases in your server, can you point me at the code please?

The code is here: https://github.com/MSheshera/lace_deployment/

And some preliminary documentation for the API: https://retrievalapp-zso5o2q47q-uc.a.run.app/docs#

hmm, @RonLek I presume there is some part of your code where you make an API call to add keyphrases if a reviewer adds it. Can you point Melisa to that as well? 

But some more context, the initial set of keyphrases for a reviewers profile will be computed offline and added to the API by a separate cron job running on Unity. Implementing this is a todo as of now, but for now the backend has a hardcoded set of reviewers and their profiles to help debug the front end.

components/webfield/ExpertiseConsole.js
RonLek: Thanks Melisa. Refactored it to be a parameter which can be passed from [here](https://github.com/RonLek/openreview-web/blob/feat/expertise-ui/pages/invitation/index.js#L92). 

Please let me know if you'd like this to be a part of the `webComponentProps` instead, although I'm not sure how [componentObj](https://github.com/RonLek/openreview-web/blob/feat/expertise-ui/pages/invitation/index.js#L64) should be modified only for ExpertiseConsole while still keeping the code generalized.

RonLek:
> hmm, @RonLek I presume there is some part of your code where you make an API call to add keyphrases if a reviewer adds it. Can you point Melisa to that as well?

This happens [here](https://github.com/RonLek/openreview-web/blob/feat/expertise-ui/components/ControllableExpertiseSelector.js#L76-L79) and is triggered if `includedKeyphrases` is modified (when a user includes/excludes/adds a keyphrase). New keyphrases are included by default.

The [backend](https://github.com/MSheshera/lace_deployment/blob/main/retrieval_app/app.py#L20) infers when it's an edit (add/delete) operation (`selected_kps` are not a subset of previously saved keyphrases) vs when it's just an include/exclude operation.

Docstring:
```
- if selected_profile_kps are NOT passed: Return recommendations computed based on
        the last saved/initial profile keyphrases for the user_id passed
- if selected_profile_kps are passed and they DONT represent an edit operation (add/delete kps):
        Lookup the profile kps latest representations and rank with them.
- if selected_profile_kps are passed and they DO represent an edit operation (add/delete kps):
        infer when the selected_kps represent an edit operation (they are not a subset of the current KPs),
        compute kp-to-paper assignment, recompute recommendations
```

melisabok:
> This happens [here](https://github.com/RonLek/openreview-web/blob/feat/expertise-ui/components/ControllableExpertiseSelector.js#L76-L79) and is triggered if includedKeyphrases is modified (when a user includes/excludes/adds a keyphrase). New keyphrases are included by default.

I don't think I understand, that code is doing a fetch operation and not a post. Where do you send the data from the UI to your server to store it in your database?

RonLek:
> I don't think I understand, that code is doing a fetch operation and not a post.

I see. The same GET endpoint is used to update the session variables. I agree this should be managed by a separate POST endpoint on the backend (cc: @MSheshera). Also, fyi currently there is no database on the backend. All user variables are session variables (and are lost if the server is stopped). We're currently looking to use Redis on GCP for this.

> Where do you send the data from the UI to your server to store it in your database?

The `user_data.update_profile` is used to update within the same [GET endpoint handler](https://github.com/MSheshera/lace_deployment/blob/main/retrieval_app/app.py#L83-L87). Session variables are updated [here](https://github.com/MSheshera/lace_deployment/blob/main/retrieval_app/rec_model.py#L225-L229) or [here](https://github.com/MSheshera/lace_deployment/blob/main/retrieval_app/rec_model.py#L234-L240) depending on if a recompute is needed. The plan is to write them in our db (Redis) instead of overwriting these `*_saved*` variables [here](https://github.com/MSheshera/lace_deployment/blob/main/retrieval_app/rec_model.py#L52-L55). As of this PR, clicking "Save Expertise" just [displays an alert](https://github.com/RonLek/openreview-web/blob/feat/expertise-ui/components/ControllableExpertiseSelector.js#L108-L112).

melisabok:
Thanks for the details but if we are going to send sensitive data to your server this must be a POST operation and there should be authentication. 

> All user variables are session variables (and are lost if the server is stopped). We're currently looking to use Redis on GCP for this.

What does it mean? if the server is down, the expertise selected by the users is lost?

melisabok:
I can not access to https://github.com/MSheshera/lace_deployment, could you send me the invitation again?

MSheshera:
> Thanks for the details but if we are going to send sensitive data to your server this must be a POST operation and there should be authentication.

Yes of course, I'll make the change on the backend. 

> > All user variables are session variables (and are lost if the server is stopped). We're currently looking to use Redis on GCP for this.
> 
> What does it mean? if the server is down, the expertise selected by the users is lost?

This is our current implementation. We will update the backend to use Redis to retain all the reviewer data and the changes induced by their interactions across sessions. However, I'd like to understand this: The current backend was mainly built to support the development of the front end and several of the changes being discussed above are being worked on at the moment, do you prefer for us to make these changes before this PR is accepted? 

@melisabok: I also invited you to the backend repo. https://github.com/MSheshera/lace_deployment/invitations

components/ControllableExpertiseSelector.js
melisabok: Authentication must be supported if we are going to ask the OpenReview users to send and receive sensitive data from your servers.

In order to call to a server with a different domain: from openreview.net to your sheshera.server.com, your server must accept calls from a different domain for security reasons. More info here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

melisabok:
> do you prefer for us to make these changes before this PR is accepted?

The UI must work with a server that support authentication, POST operation to send data from the UI to the server and persist the user changes in a data store that can be recovered if the server is restarted. 
Without this we can not offer this experimental UI to the venues. The data must be protected.

MSheshera:
> > do you prefer for us to make these changes before this PR is accepted?
> 
> The UI must work with a server that support authentication, POST operation to send data from the UI to the server and persist the user changes in a data store that can be recovered if the server is restarted. Without this we can not offer this experimental UI to the venues. The data must be protected.

Got it! We'll prioritize implementing those.

components/ControllableExpertiseSelector.js
```diff
@@ -0,0 +1,300 @@
...
+
+export default function ControllableExpertiseSelector() {
+  const [keyphrases, setKeyphrases] = useState({ active: [], inactive: [] })
+  const [includedKeyphrases, setIncludedKeyphrases] = useState([])
+  const [clickedKeyphrase, setClickedKeyphrase] = useState('')
+  const [recommendedPapers, setRecommendedPapers] = useState([])
+  const [activePapers, setActivePapers] = useState({})
+  const [keyphraseColors, setKeyphraseColors] = useState({})
+  const [paperWeights, setPaperWeights] = useState({})
+  const { user } = useUser()
```
carlosmondra, line 16: I was mistaken about the accessToken. You do need to pass it for requests to the API. You can retrieve it like so:
```suggestion
  const { user, accessToken } = useUser()
```

components/ControllableExpertiseSelector.js
```diff
@@ -0,0 +1,300 @@
...
+
+  useEffect(() => {
+    async function fetchKeyphrases() {
+      try {
+        const keyphraseResponse = await fetch(
+          `https://devapi2.openreview.net/expertise/controllable/get_recommendations/mccallum`
+        )
```
carlosmondra, line 23: In this case, I think you should import `api-client.js`. This file will figure out the base url depending on the environment. The following is my guess of how it should be, but please make sure you check.
```suggestion
        const keyphraseResponse = await api.get('/expertise/controllable/get_recommendations/mccallum', {}, { accessToken })
```

RonLek: Thanks. Updated this after local testing.

components/ControllableExpertiseSelector.js
RonLek: Thanks added.

PR #1951, opened 2024-05-13
Fix/ Forum page: reply note invitation buttons

zbialecki:
Dont show invitation buttons on replies where note readers don't match. Fixes "Readers do not match" error some users were getting when clicking the Chat invitation button.

### DIFF
--- a/components/forum/ForumReply.js
+++ b/components/forum/ForumReply.js
@@ -13,6 +13,17 @@ import Icon from '../Icon'
 import { getInvitationColors } from '../../lib/forum-utils'
 import { prettyId, prettyInvitationId, forumDate, buildNoteTitle } from '../../lib/utils'
 
+function scrollToNote(noteId, showEditor) {
+  const el = document.querySelector(
+    `.note[data-id="${noteId}"]${showEditor ? ' .invitation-buttons' : ''}`
+  )
+  if (!el) return
+
+  const navBarHeight = 63
+  const y = el.getBoundingClientRect().top + window.scrollY - navBarHeight
+  window.scrollTo({ top: y, behavior: 'smooth' })
+}
+
 export default function ForumReply({
   note,
   replies,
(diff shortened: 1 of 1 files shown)


components/forum/ForumReply.js
```diff
@@ -39,20 +50,20 @@ export default function ForumReply({
...
+  if (!note.ddate) {
+    replyInvitations = note.replyInvitations?.filter((inv) => {
+      const invitationReaders = Array.isArray(inv.edit?.note?.readers)
+        ? inv.edit?.note?.readers
+        : inv.edit?.note?.readers?.param?.const
+      if (!invitationReaders) return true
+
+      return (
+        !excludedInvitations?.includes(inv.id) &&
+        (!invitationReaders.includes('everyone') || note.readers.includes('everyone')) &&
```
melisabok, line 63: shouldn't the second condition be `invitationReaders.includes('everyone') && note.readers.includes('everyone')`? if the note contains everyone then the invitation should contain everyone too?

zbialecki: that would mean that the only invitations that are shown are the ones that are readable to everyone.

zbialecki: OK, I think I fixed the condition now. In order to be shown as a reply invitation it must be:
Not on the exclude list 
AND 
either the note is readable by everyone 
or the invitation has readers that aren't const 
or the invitation readers are a subset of the note readers

melisabok: Looks good to me.

melisabok (approved):

melisabok:
@xkopenreview can you review this one so we can deploy it today with the other chat PR?

components/forum/ForumReply.js
```diff
@@ -39,20 +50,21 @@ export default function ForumReply({
...
+    replyInvitations = note.replyInvitations?.filter((inv) => {
+      if (!excludedInvitations || excludedInvitations.includes(inv.id)) return false
+
+      const invitationReaders = Array.isArray(inv.edit?.note?.readers)
+        ? inv.edit?.note?.readers
+        : inv.edit?.note?.readers?.param?.const
+      return (
+        noteReadableByEveryone ||
+        !invitationReaders ||
+        invitationReaders.every((reader) => note.readers.includes(reader))
```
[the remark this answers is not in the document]
melisabok: I think when the invitation has a member that has a dollar sign we shouldn't apply this filter and show the button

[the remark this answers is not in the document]
melisabok: this should fixed it: https://github.com/openreview/openreview-web/pull/2022

PR #1952, opened 2024-05-13
Feature/ Workflow invitations

celestemartinez:
I have started hardcoding some of the workflow invitations we have so we can have an idea what this will look like. 
I tried using `PaginatedList` but I am not sure how to make it work with the indented "sub-invitations". This is what it looks like now:

Screen Shot 2024-05-13 at 12 29 58 PM

The `Edit` link should actually open a form that shows the fields that can be edited. 

To check, you can run the `test_venue_configuration.py` test here: https://github.com/openreview/openreview-py/pull/2118

### DIFF
--- a/components/group/GroupEditor.js
+++ b/components/group/GroupEditor.js
@@ -6,6 +6,7 @@ import GroupRelatedInvitations from './GroupRelatedInvitations'
 import GroupUICode from './GroupUICode'
 import GroupContent from './GroupContent'
 import GroupContentScripts from './GroupContentScripts'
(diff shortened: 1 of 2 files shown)


melisabok:
> what does it mean?

It supports a JSON type following the content schema, the UI should render the code editor like we do in the invitation editor.

melisabok:
let's discuss it in a call.

melisabok:
If the API still support it then the UI should do it too.

melisabok:
@xkopenreview should we close this PR?

PR #1953, opened 2024-05-13
Fix/ Error in profile auto merge test

xkopenreview:
this pr should fix profile auto merge test due to api behavior change

previously a user (emailA) can request for merge with another profile (emailB) without saving emailB to profile
but now it requires the user to first save emailB to profile otherwise /mergelink call will fail with invalid token

so this pr changed to save the profile to add userF.email to userA's profile


zbialecki (approved):
Thanks

PR #1924, opened 2024-04-30
Fix/ AC Console - remove hardcoded role names

xkopenreview:
this pr should remove hardcode role names from AC console

PR #1974, opened 2024-06-07
Fix/AC console - add default value

xkopenreview:
default value of reviewer name and anon reviewer name are removed from AC console in #1924 
but there are still some active venues created using the old template before https://github.com/openreview/openreview-py/pull/2128

this pr should keep those AC consoles continue working

changes in this pr should be reverted once those venues are over


melisabok (approved):

melisabok:
Thanks Xukun!

PR #1979, opened 2024-06-10
Fix/Message Reviewers modal - fix grouping from email to id

xkopenreview:
currently the message reviewers modal is grouping the messages by preferredEmail to show a count of messages that will be sent to the user

when the emails are masked, the display will be wrong because emails from the same domain are grouped together (email sending is still correct )

this pr should change the grouping from email to id
this pr should also add some tests

### DIFF
--- a/components/webfield/MessageReviewersModal.js
+++ b/components/webfield/MessageReviewersModal.js
@@ -36,7 +36,7 @@ const MessageReviewersModal = ({
     0
   )
   const primaryButtonText = currentStep === 1 ? 'Next' : 'Confirm & Send Messages'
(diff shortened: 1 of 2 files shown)


melisabok (approved):

enrubio (approved):
Looks great! I tested this with the AAAI tests and hid the emails:

enrubio:
@melisabok does this mean the change I made in openreview/openreview-api/pull/468 should be removed?

melisabok:
> @melisabok does this mean the change I made in https://github.com/openreview/openreview-api/pull/468 should be removed?

No need to remove it. 

We are going to disable the role validation soon and nobody will have access to the full email addresses listed in the profile.

PR #1981, opened 2024-06-11
Fix/Moderation - load only v2 config note

xkopenreview:
UI changes for https://github.com/openreview/openreview-api-v1/pull/2990

this pr should change the moderation page to load/post only v2 config note
also renamed configNoteV2 to configNote


enrubio (approved):
Looks good

racheljsmart (approved):

PR #1986, opened 2024-06-13
Feature/ Only allow certain groups to use non-component webfield

xkopenreview:
this pr should check for group domain and only allowed groups can execute non-component webfield code

however this means we won't be able to retire webfield


pages/group/index.js
```diff
@@ -185,11 +185,19 @@ return {
     domainGroup = group
   }
 
+  let webfieldCode = null
+  const jsAllowedGroups = ['TMLR']
```
melisabok, line 189: looks good!

I will create a group to store the allowed groups as members

PR #1987, opened 2024-06-14
Remove hardcoded names in openreview-web

enrubio:
This PR removes the rest of the hardcoded names in:
- All consoles (Authors, Ethics, Reviewer, AC, SAC, PC)
- Util functions (`prettyId`, `getGroupId`, `prettyTasksInvitationId`)


melisabok:
@enrubio there are some conflicts here.

components/webfield/ProgramChairConsole/Overview.js
```diff
@@ -954,7 +983,7 @@ const DescriptionTimelineOtherConfigRow = ({
                       scoresName
                     )}
                   >
-                    Reviewer Bids
+                    {prettyField(singularReviewerName)} Bids
```
melisabok, line 986: I'm not sure if we should show "Reviewers" everywhere instead of trying ti get the singular reviewer name

components/webfield/ProgramChairConsole/Overview.js
```diff
@@ -458,8 +481,8 @@ const MetaReviewStatsRow = ({ pcConsoleData }) => {
           }
         />
         <StatContainer
-          title="AC Meta-Review Progress"
-          hint="% of area chairs who have completed meta reviews for all their assigned papers"
+          title={`${prettyField(singularAreaChairName)} Meta-Review Progress`}
```
melisabok, line 484: do we have a property for the meta review name?

components/webfield/ProgramChairConsole/Overview.js
enrubio: Yeah I realized that we don't need to make _all_ the replacements. But for this specifically, this is what the overview tab looks like. I think it looks good, but I can change it back in some areas:

melisabok: Looks good!

Right, for AAAI the Program Committee role is not updated to a singular version, that is why I think we should use `reviewers_name` everywhere.

enrubio: Ooh, so what about the default role names like reviewers, area chairs etc. that are plural by default?

melisabok: Showing "Reviewers Recruitment" is it too bad?

enrubio: I think it makes sense in some places to keep the singular name, but I can definitely remove it in other places.

components/webfield/ProgramChairConsole/Overview.js
enrubio: This is fixed

components/webfield/ProgramChairConsole/Overview.js
enrubio: @melisabok I removed the use of singular names in some components, but I think it makes sense to keep it in the Overview tab, otherwise the headings read weirdly.

The only other place it's used is in the tabs, this is what it looks like:

Screenshot of PC Console

melisabok:
More places to remove the hardcoded roles:

https://github.com/openreview/openreview-web/blob/master/components/webfield/ProgramChairConsole/PaperStatus.js#L120

https://github.com/openreview/openreview-web/blob/master/components/webfield/ProgramChairConsole/PaperStatus.js#L128

https://github.com/openreview/openreview-web/blob/master/components/webfield/ProgramChairConsole/PaperStatus.js#L139

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -117,15 +119,15 @@ const PaperRow = ({
           referrerUrl={referrerUrl}
           shortPhrase={shortPhrase}
           submissionName={submissionName}
-          reviewerAssignmentUrl={getManualAssignmentUrl('Reviewers')}
+          reviewerAssignmentUrl={getManualAssignmentUrl(`${prettyField(reviewerName)}`)}
```
melisabok, line 122: I don't think you need to do prettyField here because the function is using the param to search the assignment invitation.

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -117,15 +119,15 @@ const PaperRow = ({
...
         />
       </td>
       {!noteContentField && areaChairsId && (
         <td>
           <ProgramChairConsolePaperAreaChairProgress
             rowData={rowData}
             referrerUrl={referrerUrl}
-            areaChairAssignmentUrl={getManualAssignmentUrl('Area_Chairs')}
+            areaChairAssignmentUrl={getManualAssignmentUrl(`${prettyField(areaChairName)}`)}
```
melisabok, line 130: same here

components/webfield/ProgramChairConsole/PaperStatus.js
```diff
@@ -136,7 +138,7 @@ const PaperRow = ({
...
```
melisabok, line 141: same here

components/webfield/ProgramChairConsole/SeniorAreaChairStatus.js
```diff
@@ -48,8 +48,136 @@ const SeniorAreaChairStatusRow = ({ rowData }) => (
   </tr>
 )
 
-const SeniorAreaChairStatus = ({ pcConsoleData, loadSacAcInfo }) => {
-  const { seniorAreaChairName, areaChairName } = useContext(WebFieldContext)
+const SeniorAreaChairStatusRowForDirectPaperAssignment = ({
```
melisabok, line 51: why do you have these changes in this PR?

enrubio: I'm not sure what you mean, I just merged master and resolved the conflicts. When I look at all the files it doesn't show that I made that change

enrubio: I could be wrong but what I think you're looking at is the merge commit?

melisabok: you are right, I don't see those changes in the PR now.

components/webfield/ProgramChairConsole/PaperStatus.js
enrubio: Fixed

components/webfield/ProgramChairConsole/PaperStatus.js
enrubio: Fixed

lib/utils.js
```diff
@@ -1272,36 +1240,11 @@ export function classNames(...classes) {
...
-
-  const groupSpecifier = invitationId.split('/-/')[0].split('/').pop()
-  if (entities.includes(groupSpecifier)) {
-    entityStr = groupSpecifier.replace(/_/g, ' ').slice(0, -1)
-  }
-
-  return `${paperStr} ${entityStr} ${prettyInvitationId(invitationId)}`
+export function prettyTasksInvitationId(invitationId, domain) {
+  return invitationId
+    .replace(`${domain}`, '')
```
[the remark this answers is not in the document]
melisabok: this is to remove the first part of the invitation id that we don't want to print, why do you say it is redudant?

I wouldn't care about the API 1 invitations, they should be all expired now.

components/webfield/ProgramChairConsole/Overview.js
```diff
@@ -773,7 +809,7 @@ const DescriptionTimelineOtherConfigRow = ({
           <h4>Description:</h4>
           <p>
             <span>
-              {`Author And Reviewer Anonymity: ${requestFormContent?.['Author and Reviewer Anonymity']}`}
+              {`Author And ${prettyField(singularReviewerName)} Anonymity: ${requestFormContent?.['Author and Reviewer Anonymity']}`}
```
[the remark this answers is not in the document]
enrubio: It will not, it will still say `Author and Reviewer Anonymity`. I can change it back.

unitTests/AreaChairConsoleMenuBar.test.js
```diff
@@ -111,7 +111,7 @@ describe('AreaChairConsoleMenuBar', () => {
       areaChairName: 'Senior_Program_Committee',
     }
     render(<AreaChairConsoleMenuBar {...props} />)
-    expect(baseMenuBarProps.propertiesAllowed).toHaveProperty('Reviewers')
+    expect(baseMenuBarProps.propertiesAllowed).toHaveProperty('reviewers')
```
melisabok, line 114: why is this lowercase now?

unitTests/ConsoleTaskList.test.js
```diff
@@ -100,6 +100,7 @@ describe('ConsoleTaskList', () => {
     noteInvitations = [
       {
         id: `${venueId}/${submissionName}5/-/CommentOne`,
+        domain: `${venueId}`,
```
melisabok, line 103: no need to use `` here, can you just pass `venueId`?

unitTests/AreaChairConsoleMenuBar.test.js
enrubio: I think this relates to the query search modal. All the properties usually use camelcase, but some components weren't formatted correctly. I made the change [here](https://github.com/openreview/openreview-web/pull/1987/commits/0791e8b2b57ea302e45ac0e717c98493255ac468)

melisabok (approved):

PR #1990, opened 2024-06-17
fix: hasInstitutionEmail not set to true when signing up with an institutional email

akshatchhabra:
These changes fix the issue of institutional email not confirmed message appearing when signing up using an institutional email.

### DIFF
--- a/components/profile/EmailsSection.js
+++ b/components/profile/EmailsSection.js
@@ -129,8 +129,8 @@ const EmailsSection = ({
     profileEmails?.map((p) => ({ ...p, key: nanoid(), isValid: true })) ?? []
   )
   const { accessToken } = useUser()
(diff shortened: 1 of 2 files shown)


melisabok (approved):

akshatchhabra:
> i think the issue is caused by institutionDomains being empty when loaded. right now it will briefly show warning when institutionDomains is empty
> 
> i think the logic can be adjusted so that the warning is shown when there's no institution email (in emails) after institutionDomains value is populated
> 
> so that instead of show warning then disappear it becomes not to show warning then show warning if there's no institution email

Does this mean that we do not show the warning message at first(even if a user has no institutional emails in their profile), but when a user confirms a non-institutional email, we show the warning?

PR #1997, opened 2024-06-18
Fix/ PC Console - Show Correct Info when SAC is assigned to papers directly

xkopenreview:
pc console assumes that only ACs are assigned to SAC
so when papers are assigned to SAC directly (loaded by edge), it only shows some note id in SAC status tab.

this pr should add the following new configs to PC console

- **sacDirectPaperAssignment** - when set to true will show tab content and menu bar similar to ac status tab
- **sacStatuspropertiesAllowed** - overwrite the query search properties of sac status tab (has no effect if sacDirectPaperAssignment is false)
- **messageSeniorAreaChairsInvitationId** - the invitation id for sending messages to sac
- **sacEmailFuncs** - custom message options (array of object with fields label and filterFunc)

when the value is true, SAC console shows paper related data similar to what AC status tab is showing.

### DIFF
--- a/components/webfield/ProgramChairConsole.js
+++ b/components/webfield/ProgramChairConsole.js
@@ -71,6 +71,7 @@ const ProgramChairConsole = ({ appContext, extraTabs = [] }) => {
     emailReplyTo,
     reviewerEmailFuncs,
     acEmailFuncs,
+    sacEmailFuncs,
     submissionContentFields = [],
     sacDirectPaperAssignment,
     propertiesAllowed,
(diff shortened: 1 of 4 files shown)


haroldrubio:
Yeah that'd be great if you could include that, with the same custom emailing options that the reviewers/AC tabs have

melisabok:
> messageSeniorAreaChairsInvitationId

Given this is the PC console, can I pass the meta invitation here?

melisabok:
it seems that using the meta invitation to send emails works fine.

melisabok (approved):
the changes look good! it works with these changes: https://github.com/openreview/openreview-py/pull/2212

PR #1998, opened 2024-06-19
Fix/ Submission list - render Mathjax formula

xkopenreview:
this pr should fix the issue that the formulas are not rendered in group homepage (submission list)

submission list is the only place that may show formula


celestemartinez (approved):
Looks great! Thanks Xukun for the fix.

melisabok (approved):

PR #1999, opened 2024-06-20
format message invitation using paper number

melisabok:
TMLR reminder link doesn't work because the message invitation is not formatted correctly.


melisabok:
Can you try it from the action editor console?

PR #2007, opened 2024-06-27
Fix/ Display odate in submission forum UI

### LINKED ISSUE #2003: Display odate in submission forum UI
The goal is to display the odate in the submission forum near "published" and "last modified". There may not be room horizontally but perhaps we can try showing "Public as of (odate)" below "published" or as a tooltip when hovering over everyone.

I think it makes sense to keep odate with the other dates.

xkopenreview:
this pr should fix #2003 

cause is explained in issue

### DIFF
--- a/components/forum/Forum.js
+++ b/components/forum/Forum.js
@@ -841,6 +841,7 @@ export default function Forum({
     numRepliesVisible.current = numVisible
 
     typesetMathJax()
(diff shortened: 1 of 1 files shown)


melisabok:
can you share a screenshot?

melisabok (approved):

PR #2016, opened 2024-07-03
Feature/ Profile improvements

### LINKED ISSUE #2014: Profile improvements
This a list of features request received from the CVPR Program Chairs and discussed with Andrew:

Public view: 

- Show the department name(if present) before the institution name in the history record:

Professor Department of Computer Science, University of Massachusetts Amherst

- Below the profile name show the title, department name, institution name for all the current positions:

Andrew McCallum
Full Professor, Department of Computer Science, University of Massachusetts Amherst
Director, Department of Computer Science, University of Massachusetts Amherst
Joined September 2016

Edit view:

- Make the country and city mandatory for the current positions.

xkopenreview:
this pr should fix #2014 

the profile page is 100px wider to make room for department

### DIFF
--- a/components/profile/BasicProfileView.js
+++ b/components/profile/BasicProfileView.js
@@ -59,6 +59,12 @@ const ProfileHistory = ({ history }) => (
       <strong>{history.position}</strong>
     </div>
     <div className="institution">
+      {history.institution.department && (
+        <span>
+          {history.institution.department}
+          {', '}
+        </span>
+      )}
       {history.institution.name}{' '}
       {history.institution.domain && <small>{`(${history.institution.domain})`}</small>}
       {(history.institution.city ||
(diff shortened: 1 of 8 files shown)


melisabok:
I think we should put the department name in this section too:

Screen Shot 2024-07-08 at 5 22 53 PM

melisabok:
>is it ok to show the error message if it's an old history but does not contain other missing fields such as position

It is ok for me. I don't want current profiles to start having errors because some of their previous history doesn't have a country.

melisabok (approved):

melisabok:
> missing api changes to make country/region and city mandatory

I think this can be deployed without this change, I'm worried that requesting country and region and city is too much and we should ask for country only, what do you think? @carlosmondra @xkopenreview

melisabok:
ok, can we make only the country field mandatory?

PR #2023, opened 2024-07-10
Feature/ PC/SAC/AC Console - Add plagiarism report link to paper status note summary

xkopenreview:
this pr contains ui changes for https://github.com/openreview/openreview-api/pull/487

which should load all ithenticate edges in pc/sac/ac console and show the weight in note summary in paper status tab
when user click on the weight, it should call /ithenticate/viewer-url to get link to ithenticate report and open it in a new tab

it adds a new config
>ithenticateInvitationId

to pc/sac/ac console, the value should be the invitation id which is used to create the ithenticate edges


melisabok:
I wonder if we can use the style as the venue id and show it below that before "Show Details"

Screen Shot 2024-07-11 at 6 27 20 PM

melisabok (approved):

PR #2026, opened 2024-07-18
Feature: Add anonymousID to Reviewer Status Tab in PC Console

haroldrubio:
Add anonymous ID to reviewer data in Reviewer Status Tab


haroldrubio:
Here is the use case:
https://github.com/openreview/openreview-py/pull/2509

We need to send reminders to reviewers to reply to an invitation that is submitted before the official review is submitted. The point of this forum reply is to make sure the reviewer will be available to submit a review later on

melisabok:
Can you share a screenshot showing the change?

melisabok:
ohhh I see, this is being used to post the message, it is not a change in the UI.

melisabok (approved):

PR #2029, opened 2024-07-24
Feature/ Webfields - add api.get to webfields

xkopenreview:
this pr should enable
> get

in webfields
which is a wrapper of api.get with access token

the use is similar to api.get except that there's no need to pass access token

in order to use this function, the webfield config will need to be changed.
below is an example of a PC console config (part) which loads the domain group and then use the loaded domain group to populate other configs instead of the domain object which is already available
```javascript
// Webfield component
return get('/groups', { id: domain.id }).then(result => {
  const domainObject = result.groups[0]
  return {
      component: 'ProgramChairConsole',
      version: 1,
      properties: {
        header: {
          title: 'Program Chairs Console'
        },
        apiVersion: 2,
        venueId: domainObject.id,
        areaChairsId: domainObject.content.area_chairs_id?.value,
        seniorAreaChairsId: domainObject.content.senior_area_chairs_id?.value,
        ...
        ...
``` 
with above change the pc console should load as usual without any change


carlosmondra:
Is this change backward compatible?

carlosmondra (approved):

PR #2031, opened 2024-07-25
Fix/PC console - error when paperX/reviewers/submitted group is added to paperY/reviewers group

xkopenreview:
an error will be thrown in

anonymousId: getIndentifierFromGroup(anonymizedGroup, anonReviewerName)

when anonymizedGroup is undefined
this happens when the paperX/reviewers/submitted is added as a member of paperY/reviewers group

fix is similar to #1617


melisabok (approved):

haroldrubio (approved):
The fix works in my environment

PR #2032, opened 2024-07-25
Feature/ Edge browser - remove emails and add copy email link

xkopenreview:
this pr contains changes from  [f468fd6](https://github.com/openreview/openreview-web/pull/2000/commits/f468fd65668f42912b855beef86c707baf64f3a6) in #2000

this pr should remove email display from edge browser and add copy email link in profile entity when
>preferredEmailInvitationId

query param is present which is the invitation to create preferred email edges

if 
>preferredEmailInvitationId

query param is not present, it should show the email get in profile as before

### DIFF
--- a/components/browser/Column.js
+++ b/components/browser/Column.js
@@ -343,6 +343,7 @@ export default function Column(props) {
               id: headOrTailId,
               content: {
                 name: { fullname: prettyId(headOrTailId) },
+                email: headOrTailId,
                 title: '',
                 expertise: [],
                 isInvitedProfile: true,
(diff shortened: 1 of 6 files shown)


melisabok:
Do we need to add this parameter `preferredEmailInvitationId` to edge browser url?

If it is not present, what does it happen?

melisabok:
> there's no copy email link

does it show the obfuscated email?

I would like to see no change for venues that don't have this invitation enabled.

melisabok (approved):

PR #2039, opened 2024-07-30
adjust login/redirect as needed for consoles

pmandler-umass:
The previous redirection for logging out resulted in going to the login screen with an odd message.  The default logout behavior sends the user to the Home Screen which makes more sense.


pmandler-umass:
Remove redirection for logging out because the default behavior is better.

enrubio (approved):
I tested out all consoles and this looks good to me, I like that it redirects to the home page

Issue #2042, opened 2024-07-31
Consoles: limit the number of reminders sent


melisabok:
ARR organizers are sending reminders from the UI and sometimes they are more than 1k emails, for example:

```
A total of 4994 reminder emails will be sent to the following reviewers:
```

so they get an API unavailable error when the emails are sent, my guess is because the browser can not handle 4994 requests. 

Some suggestions:

- Limit the number of emails to be sent from the UI and show an error is that is reached (suggest to use Python instead)
- When sending reminders from the reviewers's tab, we can send one message and 4994 recipients. Is is possible?

PR #2045, opened 2024-08-01
Fix/ Message Reviewers - batch every 1000 selected notes

xkopenreview:
currently the reviewer reminder api calls can exhaust browser resources when pc select many papers.
this pr should group every 1000 selected papers and chain the calls so that it doesn't fail

this pr should also change the send button to spinner button as it can take a while to send those messages

### DIFF
--- a/components/BasicModal.js
+++ b/components/BasicModal.js
@@ -1,6 +1,7 @@
 /* globals $: false */
 
 import { useEffect, useRef } from 'react'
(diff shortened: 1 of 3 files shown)


melisabok:
It fixes https://github.com/openreview/openreview-web/issues/2042

melisabok:
> do you think it's not enough?

yes, let's do that.

melisabok (approved):

PR #2049, opened 2024-08-06
Fix/ Forum reply - display note with no visible content

xkopenreview:
this pr should allow note reply to be displayed as [empty] when there's no visible content to the user
the note has no visible content to the user because the note invitation has readers for all content fields and the user is not included.

with changes in https://github.com/openreview/openreview-api/pull/503
the user can post the note but the posted note is not visible to the user

the user can still see contents in revisions page and activities page

the user won't be able to delete the invisible note because content is not available
the user can edit the note but there's no existing value so it's like posting a new note

### DIFF
--- a/components/forum/ForumReply.js
+++ b/components/forum/ForumReply.js
@@ -490,6 +490,14 @@ function NoteContentCollapsible({
       </div>
     )
   }
+  if (!Object.keys(content ?? {}).length)
+    return (
+      <div className="note-content-container">
+        <div className="note-content">
+          <em>[empty]</em>
+        </div>
+      </div>
+    )
 
   return (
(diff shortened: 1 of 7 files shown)


melisabok (approved):

enrubio (commented):
I tested this using reviews and added readers to all review content fields. When readers of the review fields didn't contain SACs or ACs, their consoles would get this error and the Assigned Submissions tab didn't load:
`Error: loading data: Cannot read properties of undefined (reading 'review')`

I tried removing PCs as readers of the review fields and they don't have this error for some reason. PC console still loads and they can view the hidden fields.

The forum page looks good though and everything else works as described in the description:

Screenshot 2024-08-09 at 3 12 15 PM

enrubio (approved):
Consoles are fixed, looks good!

PR #2053, opened 2024-08-06
Handle console link referring to tab that doesn't exist.

pmandler-umass:
create list of valid tabs to check against.


pmandler-umass:
It is possible that the ProfileBidConsole also needs this - I was not clear on what tables are expected.

enrubio:
Yes, we had to update the tabs to remove the hardcoded role names. This is fixed already in the PC console, so once you merge master you can use it as reference for the Ethics Chair Console. You'll need to remove the hardcode from `#paper-status` and `#ethicschair-tasks`

PR #2061, opened 2024-08-15
Only show the ""Show/Hide reviewers" if there are assigned reviewers.

pmandler-umass:
Check for reviewers before adding the Show/Hide links.

Before Reviewers are assigned:PC_Console_no_reviewers

And with Reviewers:
PC_Console_with_reviewers


enrubio (approved):
Thanks for this!

PR #2063, opened 2024-08-16
Feature/ Bidding console - allow config to specify subject area field

xkopenreview:
currently the subject area dropdown search for a fixed field "subject_areas" using /notes/search route

bidding console can now specify a config 
> subjectAreasName

which is the note content field name which is used to compare with user selected subject area
when **subjectAreasName** is specified, the other config 
> subjectAreas

should be an array of objects with 2 fields: value and description
value is used to match note content field value and description is displayed in the subject areas field dropdown

the notes filtering method is also changed because not all note content fields are indexed so the filtering logic is moved inside the bidding console after getting all notes

### DIFF
--- a/components/webfield/BidConsole.js
+++ b/components/webfield/BidConsole.js
@@ -249,7 +249,9 @@ const AllSubmissionsTab = ({ bidEdges, setBidEdges, conflictIds, bidOptions }) =
           { accessToken }
         )
         setNotes(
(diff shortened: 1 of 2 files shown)


enrubio (approved):
This is awesome, thanks Xukun! Tested this with the AAAI tests and it works great

Screenshot 2024-08-16 at 3 19 22 PM

PR #2064, opened 2024-08-16
Feature/ PC Console - allow sending emails to authors

xkopenreview:
this pr should add an option in pc console paper status tab to email authors of selected papers

### DIFF
--- a/components/webfield/MessageReviewersModal.js
+++ b/components/webfield/MessageReviewersModal.js
@@ -188,7 +188,7 @@ const MessageReviewersModal = ({
         <>
           <p>
             {messageOption?.value === 'allAuthors'
-              ? `You may customize the message that will be sent to authors. You can also use {{fullname}} to personalize the recipient full name. If your message is not specific to ${submissionName.toLowerCase()}, please email from the author group.`
+              ? `You may customize the message that will be sent to authors. You can also use {{fullname}} to personalize the recipient full name. If your message is not specific to a ${submissionName.toLowerCase()}, please email from the author group.`
               : `You may customize the message that will be sent to the ${prettyField(
                   reviewerName
                 ).toLowerCase()}. In the email
--- a/components/webfield/MessageReviewersModal.js
+++ b/components/webfield/MessageReviewersModal.js
@@ -188,7 +188,7 @@ const MessageReviewersModal = ({
         <>
           <p>
             {messageOption?.value === 'allAuthors'
-              ? 'You may customize the message that will be sent to authors. You can also use {{fullname}} to personalize the recipient full name.'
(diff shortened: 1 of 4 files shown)


enrubio (commented):
Can PCs not customize the email template? Even with just `{{fullname}}`?

enrubio:
I see, it is working. Then I think we should add this as an instruction like we do for the reviewers and ACs, maybe we can just say:

`You can use {{fullname}} to personalize the recipient full name.`

I also feel PCs might use this to send messages to all authors (instead of going to the group), in which case authors would get multiple of the same email if they have more than 1 submission. Do you think this is fine? Maybe it's just up to the PCs to review the list of recipients.

enrubio:
> do you think there's a use case that pc need to notify all authors related to the papers they submitted?

I'm not sure. But I agree that it would be good to add to the instructions. For ex:

`If your message is not specific to a paper, please use the author group.`

If we do this then maybe we revert this change?:

> hide email authors option when all papers are selected

Everything else is working as expected.

components/webfield/MessageReviewersModal.js
```diff
@@ -171,15 +186,19 @@ const MessageReviewersModal = ({
       {error && <div className="alert alert-danger">{error}</div>}
       {currentStep === 1 ? (
         <>
-          <p>{`You may customize the message that will be sent to the ${prettyField(
-            reviewerName
-          ).toLowerCase()}. In the email
+          <p>
+            {messageOption?.value === 'allAuthors'
+              ? `You may customize the message that will be sent to authors. You can also use {{fullname}} to personalize the recipient full name. If your message is not specific to ${submissionName.toLowerCase()}, please email from the author group.`
```
enrubio, line 191: Just a small change:
```suggestion
              ? `You may customize the message that will be sent to authors. You can also use {{fullname}} to personalize the recipient full name. If your message is not specific to a ${submissionName.toLowerCase()}, please email from the author group.`
```

enrubio (approved):
Looks good!

carlosmondra:
If a user has two submissions what will be the difference between message 1 and message 2 or how can that difference be made in the body of the email?

enrubio:
Could we add another token for submission title or number so they know what paper to check? It could useful in the other message options/templates as well

carlosmondra:
> > If a user has two submissions, what will be the difference between message 1 and message 2 or how can that difference be made in the body of the email?
> 
> there won't be any difference between message 1 and message 2 the message reviewer modal is only replacing {{submit_review_link}} which does not apply for authors
> 
> i view it as a way to select a subset of authors from the authors group based on some submission filtering criteria

I understand and it's a feature we need, but we shouldn't send the exact same message more than once to an author. I agree with Emilia, if we can add the note number or something similar, then it's fine. Otherwise, we should only send one message per author.

enrubio:
The updates look good! Should we also add this token to the reviewer message templates? I think it could be helpful.

carlosmondra (approved):
💯

PR #2066, opened 2024-08-20
Add loading graphic for groups page

pmandler-umass:
Add a state var and LoadingSpinner while loading groups.


pmandler-umass:
Difficult to test.  I stopped openreview-api from running and did see the loading pop up right before it noticed it was not responding and put up a different error.

pmandler-umass:
@xkopenreview where would I find the network throttling setting?  I search "No throttling" and didn't find anything.

PR #2071, opened 2024-08-21
Feature/ Profile edit page - use steps to show section by section

xkopenreview:
this pr should change profile edit page to use steps so that only one section/a few sections are displayed


carlosmondra:
This is very nice! I like how it's looking.

I'm wondering a couple of things:
- Is there a reason why the UI doesn't validate the current section after you change sections? Right now, sections are validated until the user clicks on "Save Profile".
- I was also wondering if it would be possible to show all the errors at once instead of one by one. Maybe this is not necessary if the above point is "fixed".
- Point 2 has Gender as its title. I wonder if we should call it something else like "Personal Information". But let's see what others think.
image

carlosmondra:
> activate and edit looks a bit different so please take a look at both and give some feedback

What are the main differences? I wasn't able to notice them.

carlosmondra (approved):

emily-grabowski:
In the Personal links section, I think we should add some additional text:
Enter full URLs of your public profiles on other sites. All URLs should begin with http:// or https://
Add:  At least one URL is required that displays your name and email.

emily-grabowski:
I would also suggest putting the required information (Name, email, homepage, history), and moving personal information later in the workflow.

emily-grabowski:
Even if we can't validate each page, could we add asterisks to highlight required sections?

enrubio (commented):
Couple things:

- In Step 2, the subtitle says `Gender, Pronouns, Birth Year and Profile Visibility` but when you're activating your profile there is no Profile Visibility option:

Screenshot 2024-09-06 at 4 19 34 PM

- I see that when there's an error we put an X on the step. So when a user gets a duplicate homepage error, should we put an X on Step 4? To show where the error is:
Screenshot 2024-09-06 at 4 30 22 PM

emily-grabowski:
> > Even if we can't validate each page, could we add asterisks to highlight required sections?
> 
> i think for new profiles, it displays what info is required on top do you think it's not enough?

I think that would be enough, then

emily-grabowski:
> > I would also suggest putting the required information (Name, email, homepage, history), and moving personal information later in the workflow.
> 
> i think putting name,email,homepage and history all in one step will make the page so long that "save profile change"/"register" button will not be always visible
> 
> i think putting all the required info together would encourage users to leave other fields empty

Yes that might be true. I think users can also easily get overwhelmed with the amount of information that we are asking for.

racheljsmart:
Screen Shot 2024-09-10 at 3 17 46 PM
I got this error when attempting to save a profile during signup. I added a second name during the process.

enrubio:
> i think for new profiles, it displays what info is required on top
> do you think it's not enough?

I still think adding an asterisk next to the required fields is a good idea. I don't think it's clear enough just by looking at the step titles that those are the required fields. Plus we already do this in other parts of openreview and I think it's pretty standard. What do you think @xkopenreview?

carlosmondra:
Another consideration, based on [this issue](https://github.com/openreview/openreview/issues/177). Should we remove the view profile button at the top or have it save the profile before going back to the profile?

enrubio (approved):
New changes look great. I really like that we don't show the view banner anymore, it makes the UI really clear.

racheljsmart (approved):

carlosmondra (approved):
I reviewed it again. I like the changes. Thanks, Xukun!

emily-grabowski (approved):
Looks good! Thanks for working on this

PR #2072, opened 2024-08-21
Fix/ Author Console - change profiles call to load for each paper

xkopenreview:
this pr should update the profile loading logic using emails in author console
to avoid the case that multiple emails of one user are used in different papers and some papers will show the user is not active


emily-grabowski (approved):

enrubio (approved):
I recreated the issue and this fix is working as expected.

PR #2075, opened 2024-08-26
Fix/ Forum - Replyto is overwritten when editing reply in linear view

xkopenreview:
this pr should fix an issue reported by user that

when user edit a reply B to a reply A in linear view in forum page, the replyto field of reply B is changed to forum note
and it will cause reply B to be shown below forum note instead of reply A even if user switch to nested view.

this is caused by forumNote always being passed as parentNote in linear view which is wrong.


celestemartinez (approved):
Works well!

enrubio (approved):
Tested this and it is working as expected.

PR #2077, opened 2024-08-27
Fix/ Assignment stats: If the assignment is to another group, change "Papers" to group name

xkopenreview:
this pr should update title/axis names in assignment stats page

currently it's reading config note's assignment_invitation content field to get invitation.edge.head.inGroup for the head name
when inGroup does not exist (for example role to paper matching) then it defaults to "papers"

### DIFF
--- a/components/assignments/HistogramStat.js
+++ b/components/assignments/HistogramStat.js
@@ -94,7 +94,6 @@ ${customLoadInvitation ? `;${customLoadInvitation},head:ignore` : ''}\
       .attr('y', 15)
       .attr('text-anchor', 'middle')
       .text(stats.name)
-      .style('font-size', '1rem')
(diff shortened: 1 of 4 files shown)


enrubio (commented):
Is it possible to also determine the tail? For SAC-AC assignments we're using the terms “users” and “ACs” and I think that can be confusing. For example:





Long text will also get cut off:

enrubio:
> i set it to read Invitation?.edge?.tail?.param?.options?.group
> is this correct?

Yes that's correct

enrubio (approved):
It looks great! Example of SAC-AC stats page:

Screenshot 2024-09-04 at 12 59 23 PM

Screenshot 2024-09-04 at 12 59 52 PM

PR #2079, opened 2024-08-29
Change Add button handling to prevent multiple presses.

pmandler-umass:
Moved handleAddButtonClick into a useEffect. State var 'isAdding' both stops more key presses and starts the isAdding process.


pmandler-umass:
Much of the code is moved, there aren't that many changes to logic.

PR #2080, opened 2024-08-30
Feature/ Toggle button widget

xkopenreview:
this pr should implement a toggle button widget

as toggle button does not have a null/empty state
it will set value to false once it's displayed (for example the submission invitation with no default value)


enrubio (approved):
Looks good! Setting a default worked when I set it as:
`"default": "true",`
where true is a string.

Screenshot 2024-09-10 at 4 57 51 PM

celestemartinez (approved):

PR #2082, opened 2024-09-04
Fix/ Console Email Copy - not to scroll to top

xkopenreview:
this pr should fix the following issues:

1. message/error prompt causing the page to scroll to top in consoles
2. message/error prompt causing the page to scroll to top in journal (templates)
3. when user has no id the template still shows copy email (for example action editor is "No Action Editor")


carlosmondra (approved):
Works as expected

PR #2084, opened 2024-09-05
Feature/ DatePicker widget - support maximum and minimum

xkopenreview:
this pr should add support of 

- maximum
- minimum
- range

to date picker widget
the dates outside of maximum, minimum and range are disabled


enrubio:
Where do you actually set the max, min, and range?

carlosmondra (approved):
Very cool! Works as expected.

enrubio (approved):
Looks good!

PR #2085, opened 2024-09-05
Fix/PC Console - selected ids should be cleared when user perform new search

xkopenreview:
this pr should fix AC/SAC/PC console to clear selected note ids when user 
- perform a new (basic or query) search
- clear search term/query


enrubio (approved):
This is working as expected.

PR #2086, opened 2024-09-05
Fix/ Update tests and test setup

xkopenreview:
dblp validation rule is changed in https://github.com/openreview/openreview-api/pull/548

this pr should update the setup data  and error message so that tests can pass


carlosmondra:
Thanks!

PR #2089, opened 2024-09-09
Fix/Edge browser- update checkbox wording

xkopenreview:
with changes in https://github.com/openreview/openreview-web/pull/1480

it's possible that checking the filter checkbox shows more entities which contradicts with user's intuition that checking the checkbox should show less items because the label says "Only show ..."

this pr should remove "only"


carlosmondra (approved):
Thanks!

PR #2090, opened 2024-09-09
Fix/ AC console - Show correct meta review when there are multiple

xkopenreview:
currently AC console shows the first meta review because it assumes there are only 1 meta review which is not always true.

~~this pr should add a config ~~
~~>skipMetaReviewSignatureCheck~~

~~which has default value false~~
compare the signature
to decide whether to compare meta review signature so that the AC console shows the meta review of the logged in user (AC)
for the use case that multiple ACs are assigned but action is required from only 1 AC skipMetaReviewSignatureCheck should be set to true

### DIFF
--- a/app/profile/edit/LimitedStateAlert.js
+++ b/app/profile/edit/LimitedStateAlert.js
@@ -2,7 +2,6 @@ import Alert from '../../../components/Alert'
 import Icon from '../../../components/Icon'
 
 export default function LimitedStateAlert({ profile }) {
-  console.log('LimitedStateAlert', profile)
   if (profile?.state !== 'Limited') return null
 
   return (
--- a/app/search/Search.js
+++ b/app/search/Search.js
@@ -1,11 +1,9 @@
 import { useEffect, useState } from 'react'
 import { truncate } from 'lodash'
 import NoteList from '../../components/NoteList'
-import PaginationLinks from '../../components/PaginationLinks'
 import api from '../../lib/api-client'
 import useUser from '../../hooks/useUser'
 import LoadingSpinner from '../../components/LoadingSpinner'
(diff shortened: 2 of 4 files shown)


celestemartinez (approved):
Looks good! The logged in AC can see their recommendation in the AC console.

enrubio (commented):
What about the recommendation values that are shown in the SAC and PC consoles? I feel like it's weird/inconsistent to show the recommendations for some papers, but not others even when the meta review is submitted. Right now we only show them if it was posted by the AC:

SACconsole_metareview

PCconsole_metareview

The AC console does look as expected. But in the case of AAAI where there are multiple invitees to the meta review invitation but only 1 note is expected, I feel it would be confusing to show `Submit` to the AC when they actually have no action. For context, Paper 2's meta review was submitted by the SAC:

ACconsole_metareview

So I'm not sure how AAAI's workflow fits with these changes. But that's my main concern.

melisabok:
@xkopenreview there are some conflicts here

melisabok:
> for the use case that multiple ACs are assigned but action is required from only 1 AC skipMetaReviewSignatureCheck should be set to true

in the case that only one meta review is needed but there are multiple ACs, shouldn't be the signature of the note the SubmissionX/Area_Chairs group? otherwise how do you check that only one meta review is needed?

components/webfield/AreaChairConsole.js
```diff
@@ -417,6 +417,7 @@ const AreaChairConsole = ({ appContext }) => {
     extraExportColumns,
     preferredEmailInvitationId,
     ithenticateInvitationId,
+    skipMetaReviewSignatureCheck = false,
```
melisabok, line 420: is this property really needed? I don't see it being used.

PR #2091, opened 2024-09-10
Fix/ Tasks - TEMP fix for completion status of AE recommendation task

xkopenreview:
this pr is a temporary fix for:

AE recommendation task of TMLR is for author to recommend 3 AEs for each paper submitted
when the author recommend 3 AEs, the tasks is displayed as completed in author console but still pending in tasks page

the issue is caused by the edge being posted using the non paper specific AE recommendation invitation while the invitations loaded in tasks page (and author console tasks tab) are paper specific. the consequence is that the repliedEdges can't be loaded correctly in invitation details and the task status is pending

TMLR author console is working because a custom logic being added because it's not using component webfield.

this pr should serve as a temporary fix of the discrepancy between author console and tasks page

this pr assumes:
- ae recommendation invitation ends with /Action_Editors/-/Recommendation
- the venue domain is the same as venue id

### DIFF
--- a/pages/tasks.js
+++ b/pages/tasks.js
@@ -32,13 +32,28 @@ const Tasks = ({ appContext }) => {
       api
         .getCombined(
           '/invitations',
-          { ...commonParams, replyto: true, details: 'replytoNote,repliedNotes', type: 'notes'},
-          { ...commonParams, replyto: true, details: 'replytoNote,repliedNotes,repliedEdits', type: 'note' },
+          {
+            ...commonParams,
+            replyto: true,
+            details: 'replytoNote,repliedNotes',
+            type: 'notes',
+          },
+          {
+            ...commonParams,
+            replyto: true,
+            details: 'replytoNote,repliedNotes,repliedEdits',
+            type: 'note',
+          },
           commonOptions
         )
         .then(addPropertyToInvitations('noteInvitation')),
       api
-        .getCombined('/invitations', { ...commonParams, type: 'tags' }, { ...commonParams, type: 'tag' }, commonOptions)
+        .getCombined(
+          '/invitations',
+          { ...commonParams, type: 'tags' },
+          { ...commonParams, type: 'tag' },
+          commonOptions
+        )
         .then(addPropertyToInvitations('tagInvitation')),
       api
         .getCombined(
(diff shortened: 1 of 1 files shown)


pages/tasks.js
```diff
@@ -32,13 +32,28 @@ const Tasks = ({ appContext }) => {
...
           commonOptions
         )
         .then(addPropertyToInvitations('noteInvitation')),
       api
-        .getCombined('/invitations', { ...commonParams, type: 'tags' }, { ...commonParams, type: 'tag' }, commonOptions)
+        .getCombined(
+          '/invitations',
+          { ...commonParams, type: 'tags' },
+          { ...commonParams, type: 'tag' },
+          commonOptions
```
carlosmondra, line 55: Not sure why you are querying tag invitations, but you only need to pass one:
```suggestion
          '/invitations',
          { ...commonParams, type: 'tag' },
          commonOptions
```

pages/tasks.js
```diff
@@ -51,7 +66,37 @@ const Tasks = ({ appContext }) => {
...
+      .then(async (allInvitations) => {
+        const aERecommendationInvitations = allInvitations[2].filter((p) =>
+          p.id.endsWith('/Action_Editors/-/Recommendation')
+        )
+        const aERecommendationEdgesP = aERecommendationInvitations.map((p) =>
+          api
+            .get('/edges', {
+              invitation: `${p.domain}/Action_Editors/-/Recommendation`,
+              groupBy: 'head',
+              domain: p.domain,
```
carlosmondra, line 78:
```suggestion
              invitation: `${p.domain}/Action_Editors/-/Recommendation`,
              groupBy: 'head',
```

carlosmondra (approved):

pages/tasks.js
carlosmondra: Rarely and definitely not in this case, because it doesn't help reduce the amount of documents returned.

pages/tasks.js
carlosmondra: Correct, but you only need to pass singular. The API 2 supports plural to avoid breaking the UI, but it should be singular

carlosmondra: aaah I understand now. Yes, it's correct. No need to change anything.

celestemartinez (approved):

PR #2092, opened 2024-09-12
Fix/ Login page - add tooltip when email is invalid

xkopenreview:
some users overlook the "Email" label and think they can use tild id or name to login

currently the login button is only disabled

this pr should add a tooltip when the email that user entered is invalid


enrubio (approved):
Looks good!

Screenshot 2024-09-17 at 1 19 16 PM

PR #2093, opened 2024-09-13
Fix/ PC/SAC Console: Show link to edit assignments only if the invitation is active

xkopenreview:
this pr should separate the logic of auto and manual assignments to show edit assignment link

auto assignments can be deployed (assignment invitation being created) then undeployed (assignment invitation still exists but expired) and the edit assignment link should not be displayed

manual assignment has no undeployment so only check for existence of assignment invitation


celestemartinez (approved):
Looks good!

haroldrubio (approved):

PR #2095, opened 2024-09-16
Fix/ Edge Browser: support textbox input when the edit invitation is not an enum

xkopenreview:
this pr should allow a textbox in edge browser to add/edit edge

example for weight of custom max papers:
```javascript
"weight": {
    "param": {
      "input": "text",
      "maximum": 100,
      "minimum": 0
    }
  }
``` 

should show a textbox for custom max papers edge

### DIFF
--- a/components/forum/Forum.js
+++ b/components/forum/Forum.js
@@ -12,6 +12,8 @@ import groupBy from 'lodash/groupBy'
 import escapeRegExp from 'lodash/escapeRegExp'
 import List from 'rc-virtual-list'
 
+import dayjs from 'dayjs'
+import relativeTime from 'dayjs/plugin/relativeTime'
 import ForumNote from './ForumNote'
 import NoteEditor from '../NoteEditor'
 import ChatEditorForm from './ChatEditorForm'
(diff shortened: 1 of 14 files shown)


celestemartinez:
How does a default value work here? Because when I set a default value and I try to assign a reviewer that already has that number of assignments, the edge browser won't let me make the assignment, but I don't see this default value anywhere and it's not clear why I can't make the assignment. 

Also, it is not very straightforward how to change this value, and changing it is not very smooth. Not sure if this will be a good experience for PCs.

enrubio (commented):
A few things:

I think the trash icon should look the same as the Assignment one:
Screenshot 2024-09-26 at 4 31 05 PM

If you put invalid input, the spinner goes on forever with no way of editing your previous input. I think the expected behavior is to immediately edit and fix the issue. Otherwise you have to click another paper and click back, or trash the edge (if it exists):
spinner

I think we should also have some way of telling the PCs what the expected input is now that it's a textbox.

enrubio (approved):
The issues I mentioned have been fixed and I think it looks good. I was worried that the expected values/range wasn't as clear as it was with the dropdown, but I checked again and I think the validation errors are clear enough.

Example with and without a default (of 6):

cmp-text

PR #2098, opened 2024-09-17
Fix/ PC Console - separate propertiesAllowed config for AC status tab

xkopenreview:
this pr should add a config in PC console
> areaChairStatusPropertiesAllowed

which is the query search config overwrite for AC status tab

currently AC status tab reads the config "propertiesAllowed" which is for paper status tab but the data structure of paper status tab and ac status tab are different so it probably won't work.

the behavior of overwriting the default config is not changed because it's consistent with SAC status tab (direct paper assignment)


enrubio (approved):
Looks good to me, tested that the default config does get overwritten.

PR #2101, opened 2024-09-18
Feature/ Moderation - fetch urls in profile preview

xkopenreview:
this pr should have the profile preview to auto fetch urls and show an icon to indicate the status

- check: the destination server responded, doesn't mean it's a valid homepage/url of the user
- cross: the server failed to respond, maybe the url is invalid

and it serves as an indicator for the moderator

### DIFF
--- a/.testcaferc.json
+++ b/.testcaferc.json
@@ -1,5 +1,5 @@
 {
-  "browsers": "chrome",
+  "browsers": "firefox",
   "videoPath": "videos",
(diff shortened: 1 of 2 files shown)


emily-grabowski (approved):
Seems like this will be helpful!

emily-grabowski (approved):
Does this search both Bing and Google? Or one or the other

racheljsmart (approved):

PR #2103, opened 2024-09-19
Fix/ Invitation revisions page - edit to remove process function is not displayed correctly

xkopenreview:
requires some discussion about removing process/web from invitation and group

discussed with Carlos
for now the way to delete webfield/process function is to pass space


carlosmondra (approved):
In the future, we need to figure out a way to pass `{ delete: true }` from the UI.

enrubio (commented):
I tested this and it does work. Should we also show `(empty)` when the preprocess is removed? It just shows blank right now:

Screenshot 2024-10-15 at 2 49 42 PM

melisabok:
more conflicts here @xkopenreview

melisabok (approved):

PR #2106, opened 2024-09-23
Fix/Feature proposal: render invitation buttons in gray color when invitation is expired but user is writer

xkopenreview:
this pr should show expired invitations differently from other invitations in

- forum add reply buttons
- forum edit dropdown

so that pc won't be confused whether the invitation expiration is working or not


carlosmondra (approved):
I like this change! Works as expected. Thanks.

celestemartinez:
I like this change! Two comments:
1. Can we capitalize the word "Expired"?
2. Could the tooltip say that it can still be used by Program Chairs only? I would like to make this clear so PCs don't think anyone can use the button?

celestemartinez (approved):
Looks good, unless anyone has any other ideas for the wording:
Screenshot 2024-10-15 at 3 28 35 PM

carlosmondra:
> Looks good, unless anyone has any other ideas for the wording

I am OK with the wording 👍

enrubio (commented):
Tested this and it does work, but I wonder if we can remove the strike through the text? I don't think it's necessary since it's already grayed out and has a tool tip. I also think it can make it harder to read and conveys that it can't be used at all, which isn't true. For ex:

no-strike

I'm also unsure about the icon added because it doesn't relate to what the invitation does and I think it makes it confusing. I'm sure we added it because it was expired, but I think just a grayed out button + tooltip works great. Ex:

gray-no-icon-tooltip


what do you all think? @xkopenreview @celestemartinez @carlosmondra

carlosmondra:
I do like having the strike through along with the grayed text. We really don't want them to use that button unless it's absolutely necessary. Regarding the icon, I don't really have a strong opinion, so I'm fine with removing it.

carlosmondra:
@celestemartinez @enrubio and I discussed this PR in the meeting and we agreed to the following:
- Keep the strike through along with the grayed text.
- Remove the calendar icon
- The tooltip should ony say: "Expired 4 days ago"

emily-grabowski:
Since the strikethrough makes the button harder to read, could add the invitation name to the tooltip like this:

 [Invitation Name] expired 4 days ago

racheljsmart:
I second Emily's suggestion. Adding the invitation name to the tooltip helps identify the invitation and what is expired.

celestemartinez:
Looks great! One small question: this is only for the forum, right? Sometimes we get PCs confused as to why they can still see the submission invitation in the homepage after it has expired, but I don't know it this would be easy to implement there too.

carlosmondra:
> Looks great! One small question: this is only for the forum, right? Sometimes we get PCs confused as to why they can still see the submission invitation in the homepage after it has expired, but I don't know it this would be easy to implement there too.

Good point. @xkopenreview if it requires too many changes in this PR, let's do it in a separate one.

celestemartinez (approved):

carlosmondra (approved):

enrubio (approved):
Looks good! For reference:

Submission button:

Screenshot 2024-11-12 at 5 19 18 PM

Edit Submission button:

Screenshot 2024-11-12 at 5 19 48 PM

PR #2108, opened 2024-09-24
Fix/ Edge Browser: don't enable the button associated with the edit invitation if the user is not part of the tail group

xkopenreview:
this pr should disable edit edge toggle button when the following conditions are met
1. the edit invitation is traverse invitation
2. the edit/traverse invitation specifies inGroup a group
3. the profile id is not in that group

because trying to add such as edge will fail api inGroup validation

### DIFF
--- a/lib/webfield-utils.js
+++ b/lib/webfield-utils.js
@@ -237,11 +237,11 @@ ${invitation.apiVersion === 2 ? 'view2' : 'view'}.mkNoteEditor(
 export function translateInvitationMessage(invitationMessage, args) {
   if (!args || Object.values(args).some((p) => typeof p !== 'string')) return invitationMessage
   return invitationMessage.replace(/\{\{([^}]+)\}\}/g, (match, key) => {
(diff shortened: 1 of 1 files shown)


enrubio:
I'm trying to replicate the issue using the TMLR tests. What I did was:
- Pick an AE
- Create Assignment, Custom Max Papers, and Availability edges for them (via the edge browser)
- Removed them from the AE group, their edges still existing
- Impersonate an author and try to recommend that AE

The AE doesn't show up in the edge browser at all. From Melisa's comment, it looks like they should be showing with the Recommend button. Not sure if I'm misunderstanding how the issue works.

melisabok:
> I am not understanding why we are showing this user even after they have been removed from the AE group
The edge browser is building a map of all the profiles that it gets: from the AE group or from other edges (like invited reviewers). When we do a search, the map is being used. I think this makes sense in most cases where PCs are looking for reviewers. I think we should indicate the UI when to show it or not but I'm not sure how.

PR #2109, opened 2024-09-24
Fix/ Bidding Console - add enableSearch config to enable/disable search input

xkopenreview:
this pr should add a config 
>enableSearch

which has default value true in bidding console to hide the search input (when set to false)

the use case is that the user should only see papers with a score edge


celestemartinez (approved):
Seems to work well, thanks Xukun!
Reviewers can only see papers with a `Score` edge and they can't search.

Screenshot 2024-09-25 at 2 39 25 PM

PR #2110, opened 2024-09-25
Fix/ Edge browser - only show entities associated with parent on initial load

xkopenreview:
currently edge browser show items when it has traverse edge, even if it's not related to the parent entity by traverse edge (assignment) or browse edge (affinity score) at all.

this pr should filter out such items on initial load
when user perform search or max quota filtering, such items should still be displayed (refer to #1480 for rationale)

### DIFF
--- a/components/ErrorDisplay.js
+++ b/components/ErrorDisplay.js
@@ -4,7 +4,7 @@
 import CommonLayout from '../app/CommonLayout'
 import api from '../lib/api-client'
 
-const ErrorDisplay = ({ statusCode, message }) => {
+const ErrorDisplay = ({ statusCode, message, withLayout = true }) => {
   const handleLogout = async () => {
     try {
       await api.post('/logout')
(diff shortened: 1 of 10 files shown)


carlosmondra:
> when user perform search or max quota filtering, such items should still be displayed

Does it mean that if, in the case of TMLR, an AE searches for a reviewer who does not have an Edge with the submission in the first colum, the reviewer will still show up in the second column?

carlosmondra:
> > > when user perform search or max quota filtering, such items should still be displayed
> > 
> > 
> > Does it mean that if, in the case of TMLR, an AE searches for a reviewer who does not have an Edge with the submission in the first colum, the reviewer will still show up in the second column?
> 
> that's right. search for a reviewer or using quota reached checkbox

So the search is performed on all the edges just using the invitation as the filter, correct?
Is there a way to make this optional for the search? TMLR wants to avoid this from happening, but I understand why it would be a desirable feature for other venues.

carlosmondra:
> > > > > when user perform search or max quota filtering, such items should still be displayed
> > > > 
> > > > 
> > > > Does it mean that if, in the case of TMLR, an AE searches for a reviewer who does not have an Edge with the submission in the first colum, the reviewer will still show up in the second column?
> > > 
> > > 
> > > that's right. search for a reviewer or using quota reached checkbox
> > 
> > 
> > So the search is performed on all the edges just using the invitation as the filter, correct? Is there a way to make this optional for the search? TMLR wants to avoid this from happening, but I understand why it would be a desirable feature for other venues.
> 
> i think we can either has some setting in the traverse invitation to denote this or has a new query param to control whether to filter out those entities with no relation with parent

I would prefer to have a parameter in the edge browser. @celestemartinez do you have any thoughts?

PR #2111, opened 2024-09-27
Fix/ Integration test - break dependency with DBLP

xkopenreview:
this pr should fake the response xml from https://dblp.org/pid/95/7448-1
so that when dblp is down or slow integration test and deployment is not affected.

this pr only fix the dependency of web test

while then profile is saved api will make a call to check for disambiguation page
https://github.com/openreview/openreview-api/pull/549/files#diff-007dc3e2c3e35c242d17d88a2473a70e119007ea819aa74278dc2003dfcf7709R424
which may still fail the integration test

dependency with abstract extract cloud function should be fixed by https://github.com/openreview/openreview-py/pull/2361


melisabok:
@xkopenreview I don't see any changes here? should I close this PR?

PR #2117, opened 2024-10-04
Fix/ Console - support unicode in hash

xkopenreview:
when tab id has unicode and user is pasting a url with #tab, the auto switch of tabs is not working as expected due to url encoding

this pr should fix this issue by decoding the hash


enrubio (approved):
Tested and the tab switching is fixed.

PR #2120, opened 2024-10-07
Feature/ Reply preview for v2 note invitation

xkopenreview:
this pr should add a note (or reply) preview for v2 note invitation


celestemartinez:
I am not sure how helpful this is for a couple of reasons:

1. This only works for the paper-specific invitations. This means a PC would have to go to a paper-specific invitation to preview any changes, but then they would have to make the changes to the super invitation (which has no preview)
2. PCs don't know that changing the fieldname will not rename the field, but create a new field with that name (and that the previous field needs to be deleted with` {"delete": True}`). But I am not sure how we can make this clear in the UI

celestemartinez:
> > 1. This only works for the paper-specific invitations. This means a PC would have to go to a paper-specific invitation to preview any changes, but then they would have to make the changes to the super invitation (which has no preview)
> > 2. PCs don't know that changing the fieldname will not rename the field, but create a new field with that name (and that the previous field needs to be deleted with` {"delete": True}`). But I am not sure how we can make this clear in the UI
> 
> can you give an example of super invitation where preview is not shown? this function can be added to the json editor which is used in workflow invitation content field right now for invitation editor i think we still expect user to modify json directly

I was trying to use the preview for the invitation `ICLR.cc/2024/Conference/-/Official_Review`, but I could only see the preview for a paper invitation, for example `ICLR.cc/2024/Conference/Submission1/-/Official_Review`.

enrubio:
> this function can be added to the json editor which is used in workflow invitation content field

This is referring to the venue configuration changes right?

I think this is cool to have. I'm worried though about PCs seeing this, making changes directly in the invitation, then running a stage which would overwrite those changes and being confused.

melisabok (approved):

celestemartinez (approved):
Looks great!

PR #2121, opened 2024-10-08
Feature/ Add checkbox to SAC console AC status

xkopenreview:
this pr should add checkbox to ac status tab of sac console
so that sac can email individual ac


enrubio (approved):
This is working as expected.

PR #2123, opened 2024-10-09
Fix/ Edit revisions - Use code editor to show process functions

xkopenreview:
this pr should display web/process in edit history using code editor so that it has proper highlighting/line wrapping etc.

---
this pr should also fix the issue that date process can't be deleted
currently it's passing null which fails api validation, this pr should change it to pass {delete:true} instead


enrubio (commented):
This looks great, is it also possible to do this for the preprocess?

Screenshot 2024-11-07 at 3 20 57 PM

enrubio (approved):
Looks and works great. Removing the dateprocess also works now.

PR #2124, opened 2024-10-09
Fix/ Group directory show loading spinner when links are specified

xkopenreview:
changes in #2066 didn't handle the case where links are specified in config so that load child groups are not invoked
so the loading spinner is always visible

this pr should fix the logic and add the some test


carlosmondra (approved):

enrubio (approved):
This looks good

PR #2129, opened 2024-10-10
Feature/ User moderation - allow a short note to be saved with impersonated ids in the past

xkopenreview:
this pr should allow a short note to be added when impersonating a user
so that purpose of past impersonations is clear and the impersonate link can be reused


emily-grabowski (approved):
Looks like this is working as expected. Are the notes visible to other support team members? Will we also be able to see the notes of PCs who impersonate authors?

emily-grabowski:
Ok thanks for clarifying!

racheljsmart (approved):

racheljsmart:
If impersonating the same user multiple times, the note is overwritten each time. It's not really an issue, just the observed behavior.

enrubio (approved):
Changes look good.

PR #2131, opened 2024-10-14
Fix/ Profile edit page - publication not updated when user import using non-persistent url

xkopenreview:
the imported publications should update when user import some papers to reflect latest changes

this function is currently not working when user is importing using non-persistent url

when user complete importing, the dblp url will be replaced by the persistent url if it's different from the persistent url which caused by modal close handler to be registered twice which breaks the publication update logic.

this pr should check whether the modal already have a close handler so that it's not added again


emily-grabowski:
I'm testing this and when I put in the homepage URL, the papers still don't show:

Screenshot 2024-11-01 at 1 19 24 PM

emily-grabowski:
Ok I see now - it doesn't replace the URL automatically, but prevents an error when the URL is replaced

emily-grabowski:
Yes I did - I meant that the system doesn't automatically find the persistent DBLP URL based on the non-persistent URL and replace it in the profile

emily-grabowski (approved):

racheljsmart (approved):

melisabok (approved):

PR #2134, opened 2024-10-16
Update password reset success message

enrubio:
This updates the password reset success message so that users know they may not receive an email if an account is not associated with the email they entered.

We have been getting emails from users confused why they haven't received this email and this gives them one possible reason.

Ex:
Screenshot 2024-10-16 at 2 10 09 PM


emily-grabowski (approved):
The message looks great!

PR #2135, opened 2024-10-16
Feature/ Prompt Message: only the last message is displayed

xkopenreview:
this pr change the way which notification is shown
so that:
1. multiple messages can be shown (currently set to 2 max)
2. different message types can show for different duration and allow custom duration to be passed
3. messages can be closed
4. messages can show markdown/mathjax

new promptError/promptMessage/promptLogin should just replace the one used to be defined in global.js

---
current promptLogin accept a user argument which is used to determine whether to show Login or Login as a different user when editor can't get reader or signature
this apply only to v1 forum and note editor, the new readers and signature component has their own error handing so the logic is removed.

### DIFF
--- a/hooks/usePrompt.js
+++ b/hooks/usePrompt.js
@@ -18,6 +18,7 @@ export default function usePrompt() {
         ...(canClose && {
           closeIcon: <Icon name="remove" />,
         }),
+        animation: null,
       },
       (notification) => {
         setNotificationInstance(notification)
@@ -76,6 +77,7 @@ export default function usePrompt() {
           ...(canClose && {
             closeIcon: <Icon name="remove" />,
           }),
+          animation: null,
         },
         (notification) => {
           setNotificationInstance(notification)


carlosmondra:
Is it necessary to change the style of the banner? If so, what are the advantages?

carlosmondra:
> > Is it necessary to change the style of the banner?
> 
> if we still show at most 1 message then no if showing multiple using current banner style it's weird

I the changes, but I like the current style better. It matches OpenReview styling. In what case would it be better to show two messages (or more)? Maybe we can do without that.

carlosmondra:
The styling helps. I like it better now. However, the message is a bit easy to miss. The current one is very clear. My guess is that having it appear below the header will make it easier to notice. Also if the width of the message could be larger, I think it would help.

I tried the mobile view and it's not possible to close the banner from there.

Also as a separate issue: The profile page does not show up properly from the mobile view. Not sure if there is an issue already. I can create it if not.

carlosmondra:
I think I liked the close button from before better.

Is it possible to make the message background wider?

carlosmondra:
> > Is it possible to make the message background wider?
> 
> do you mean wider or taller?

Wider 😬

carlosmondra:
Changes look so much better.

I'm not sure about the `close` button. I think the X that we had in the previous banner looked better. It blended better with the style of the banner.

Also how do you configure how long a message stays on the screen?

carlosmondra (approved):
Looks great! Thanks

racheljsmart (approved):

akshatchhabra (approved):

melisabok:
Xukun, do you have some screenshots to see the messages?

melisabok:
@xkopenreview there are some conflicts here.

melisabok (approved):

melisabok:
@xkopenreview can you address this comment from Emilia? https://github.com/openreview/openreview-web/pull/2135#issuecomment-2598504059

PR #2136, opened 2024-10-16
Feature/ Reviewer console menu bar

xkopenreview:
currently the only function of reviewer console menu bar is to download pdfs

this pr should add search and sort function to reviewer console menu bar
so that reviewers can filter papers (for example by track to identify papers rejected in 1st review stage)


enrubio (commented):
Changes look good! Can we also add the `venue` property to the AC console? Seems like we have it in the other consoles (PC, SAC, Reviewer).

I also noticed that when we apply a filter with the query search, then sort, it will ignore the filter and display the full list of papers. Maybe that's something to fix in another PR.

enrubio (approved):

celestemartinez (approved):

PR #2142, opened 2024-10-24
Fix/ Moderation - remove urls fetch in profile preview

xkopenreview:
this pr should revert some of the changes in https://github.com/openreview/openreview-web/pull/2101
because the fetch call is blocked by nginx

the email search links should remain


emily-grabowski (approved):
It is working for me- the email search works but there isn't a url check
Screenshot 2024-11-06 at 10 37 42 AM

racheljsmart (approved):

PR #2143, opened 2024-10-24
Fix/ Forum page - chat display when there are a few chat messages

xkopenreview:
chat view breaks when there are only a few chat messages

this pr should fix it by removing the item height

this pr should also fix the scroll bar display issue in filter dropdown

this pr also removed forward ref of ChatReply because it's not used anywhere


carlosmondra (approved):
Thank you!

carlosmondra:
This PR fixes this issue: https://github.com/openreview/openreview/issues/239

celestemartinez (approved):

PR #2145, opened 2024-10-24
Fix/ Font loading causes text to shift on initial page load

xkopenreview:
this pr should fix the issue of text shifting caused by loading fonts from google
by using next/fonts instead

the call to fonts.googleapis.com only load the font style which contains urls to the actual font
which is a sequential two-step process and showing fallback font then showing the correct loaded font is noticeable by users.


celestemartinez (approved):

melisabok:
How do I reproduce this issue?

melisabok (approved):

Issue #2147, opened 2024-10-25
Profile Edit page - name section button not updated after adding a name


xkopenreview:
when user add a new name and click "save profile change"
the button of the newly added name is still "Remove" instead of "Make Preferred" and "Request Deletion"

PR #2148, opened 2024-10-25
Fix/ Profile Edit page - name section button not updated after adding a name

xkopenreview:
this pr should fix the issue described in #2147


emily-grabowski (approved):
Looks like its working in this PR, but I wasn't able to replicate the original issue

Screenshot 2024-11-06 at 10 37 42 AM

emily-grabowski:
Yes I was able to replicate it now - looks good

racheljsmart (approved):
I was able to replicate the issue, and the fix appears to be working.

PR #2152, opened 2024-10-29
Feature/ Forum page - show pdf page count

xkopenreview:
this pr should display image of the first page of pdf file and page count of the pdf file in forum page

looks like 
image


melisabok:
@xkopenreview what is the purpose of doing this? did someone request this feature?

PR #2161, opened 2024-10-31
Fix/ Replace pen icon with by

xkopenreview:
this pr should replace pencil icon in forum reply with by and reduce excess padding

purpose of pencil icon is unknown


carlosmondra (approved):
Works as expected. Thanks!

PR #2162, opened 2024-10-31
Fix/ Random order in Reviewer properties

xkopenreview:
when a browse invitation is ignore head invitation and it has a default value,
the default value was added to profile entity if the entity does not have such an edge.

currently these ignore head invitation default values edges are added to the end

this pr should sort the browse edges based on browse invitation so that they will be in the same sequence as in entities which has all the edges


carlosmondra (approved):
Thanks!

celestemartinez (approved):

PR #2163, opened 2024-11-01
Fix/ User moderation - moderation queue may show empty results

xkopenreview:
related to #2015 

currently when moderator 
search for profile and get only 1 result
then
approve/reject the profile

the load profiles with fail with negative offset because of the logic to go to previous page which does not appl to top queue

this pr should fix that


emily-grabowski (commented):
The PR fixes the error when I search for a profile and approve it
Screenshot 2024-11-06 at 1 44 09 PM

When the last profile in the moderation queue is moderated, I still get the offset error- is this something we can address in this PR?
Screenshot 2024-11-06 at 1 42 35 PM

emily-grabowski (approved):
Works for both cases now

racheljsmart (approved):

PR #2164, opened 2024-11-01
Feature/ Notification search

xkopenreview:
when the user is involved in many conferences, it's hard to check the notifications from a certain conference.

this pr should add the search input in notifications page so that user can search the email subject

### DIFF
--- a/components/Layout.js
+++ b/components/Layout.js
@@ -1,4 +1,6 @@
 import Head from 'next/head'
+/* eslint-disable camelcase */
+import { Noto_Sans } from 'next/font/google'
 import Nav from './Nav'
 import Banner from './Banner'
 import EditBanner from './EditBanner'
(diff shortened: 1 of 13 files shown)


carlosmondra:
So this PR makes a search on all messages using subject prefix? I feel like most people will expect to be a search over any part of the text.

carlosmondra:
> > So this PR makes a search on all messages using subject prefix? I feel like most people will expect to be a search over any part of the text. Is this related to an issue we are trying to fix?
> 
> related to [openreview/openreview#201](https://github.com/openreview/openreview/issues/201) to filter messages from certain venue

Wouldn't it be easier to use the domain keyword?

melisabok:
@xkopenreview can we do a local search?

we currently load all the messages in the notifications page, let's do the search using the loaded messages, no need to call the api

PR #2169, opened 2024-11-05
Fix/ Tasks page: Completion status of edge invitations

xkopenreview:
logic in #2091 caused the repliedEdges to be overwritten
so that bidding is not shown as completed

this pr should fix the issue by returning the original invitation if it's not AE recommendation task


carlosmondra (approved):
Hard coded stuff makes me sad

celestemartinez:
> > Hard coded stuff makes me sad
> 
> i think permanent fix requires change of invitation setup

What changes to the invitation would need to be made?

celestemartinez (approved):

emily-grabowski (approved):

PR #2173, opened 2024-11-08
Feature/ Add NotesViewer component

xkopenreview:
this is a component for users to view v2 notes array formatted like notes
access should be controlled by group readership

when there are many notes, it's better to check pagination to improve page performance

sample config:
```javascript
// Webfield component
return {
  component: 'NotesViewer',
  version: 1,
  properties: {
    header: {
      title: `Notes Viewer`,
      instructions: `Paste notes json and view it formatted`
    }
  }
}
```


carlosmondra (approved):
Thanks! Works as expected.

carlosmondra:
> changes to export CSV should not interfere with other csv download so please also test console download if possible

Tested both things. Works as expected.

carlosmondra (approved):

enrubio (approved):
Works great. Exports in all consoles seem to be working the same.

PR #2175, opened 2024-11-13
Fix/ DBLP import - support importing >1000 papers

xkopenreview:
currently there's a limit on dblp import to import at most 1000 papers

it looks like there's not really a reason to put his limit now (through max replies) with the changes implemented in the past
and it's causing issue with api performance

this pr should remove the 1000 import limit


emily-grabowski:
When I tested this I was able to import 1000 publications, then I got the message 'Import quota reached':
Screenshot 2024-11-15 at 1 18 59 PM

carlosmondra (approved):
Works as expected.
Be sure to remove the 1000 from maxReplies in http://localhost:3030/invitation/edit?id=DBLP.org/-/Record before testing

emily-grabowski (approved):

enrubio:
Does this mean allowing you to import 1000+ papers at once? When I try to test this it says 500 publications is the max allowed.

carlosmondra:
> @carlosmondra is it possible to remove the rate limiting when
> 
> > Does this mean allowing you to import 1000+ papers at once? When I try to test this it says 500 publications is the max allowed.
> 
> no. because the user will probably encounter error with rate limiting

Is this the rate limit when POSTing note edits?

PR #2179, opened 2024-11-15
Fix/ Register page - Add turnstile widget for password reset in register page

xkopenreview:
turnstile widget is only added for the reset page

this pr should add it to the register page for password reset too so that api can validate the token


carlosmondra (approved):
Works as expected. Thanks! I tested in the dev site

enrubio (approved):
Works great

PR #2181, opened 2024-11-18
Fix/ SAC Console: Fix "venue" query search property

xkopenreview:
this pr should fix
- query search by venue
- sort by venue

in sac console


emily-grabowski (commented):
The sort function works, but when I search for 'ICML 2023' in the SAC console for the ICML test, I get empty results

emily-grabowski:
Met with Xukun, and the query for venue is working correclty

emily-grabowski (approved):

enrubio (approved):
This works, thank you!

PR #2183, opened 2024-11-18
Feature / New stat in AC console and sorting mechanism

xkopenreview:
this pr should support custom sorting option in AC console

example usage to add a sort option to sort by confidence spread:
```javascript
...other configs,
sortOptions:[
      {
        label:'Confidence Spread',
        value:'confidence spread',
        getValue:"return row.reviewProgressData?.confidenceMax - row.reviewProgressData?.confidenceMin"
      }
    ]
```


carlosmondra (approved):
Thanks! It works

PR #2188, opened 2024-11-27
Feature/ Console - extract check all checkbox in different consoles

xkopenreview:
this pr should refactor duplicated select all checkbox code in
AC
SAC
PC
consoles and add some tests


emily-grabowski (approved):
Checked for SAC, PC, AC consoles and it seems to be working

enrubio (approved):
Changes look good

PR #2189, opened 2024-11-29
Feature/ Consoles - Extract console tabs into components

xkopenreview:
this pr should consolidate the tabs and active tab logic to a separate component so that related logic are not duplicated in all consoles and can be better tested


emily-grabowski (approved):
Tabs seem to be working for PC/AC/SAC consoles

components/webfield/ConsoleTabs.js
```diff
@@ -0,0 +1,61 @@
...
+      validTabIds[0]
+  )
+  const router = useRouter()
+
+  useEffect(() => {
+    if (!validTabIds.includes(activeTabId)) {
+      setActiveTabId(defaultActiveTabId)
+      updateActiveTabId?.(`#${defaultActiveTabId}`)
+      return
+    }
```
enrubio, line 19: Do we need to call `updateActiveTabId` here? I think when `setActiveTabId` is called it will trigger useEffect again with a valid (default) tab, so the parent's state would already get updated in line 20 with the default. What do you think?

enrubio (approved):
My comments were addressed and consoles are functioning as expected.

PR #2202, opened 2024-12-18
Fix/ Next.js 15 upgrade

xkopenreview:
this pr requires https://github.com/openreview/openreview-api/pull/623

this pr should tested in release mode meaning that instead of
>npm run dev

the commend to start web should be
>npm run build

followed by
>npm run start

because the behavior of caching strategy might be different in development mode and release mode


melisabok:
what about the conflicts listed here? can I start testing this or should I wait?

melisabok:
>this pr should tested in release mode

should we always use release mode or only the first time?

melisabok (approved):

carlosmondra (approved):

pages/forum.js
```diff
@@ -1,249 +0,0 @@
...
-      Router.replace(redirectUrl)
-    }
-    return {}
-  }
-
-  try {
-    const note = await api.getNoteById(
-      queryId,
-      token,
-      { trash: true, details: 'writable,presentation' },
```
xkopenreview: @melisabok i believe this is the correct file

PR #2203, opened 2024-12-18
Fix/ AC Console - Show assigned secondary AC(s)

xkopenreview:
this pr should show in AC console the secondary AC(s) contact info.
similar to assigned SAC info

it assumes the invitation to create the secondary assignment edge is 
{venue_id}/{secondary_area_chairs_name}/-/Assignment


components/webfield/AreaChairConsole.js
```diff
@@ -357,6 +365,22 @@ const AreaChairConsole = ({ appContext }) => {
         : Promise.resolve([])
       // #endregion
 
+      // #region secondary ACs
+      const secondaryAreaChairsP = secondaryAreaChairName
+        ? api
+            .get(
+              '/edges',
+              {
+                invitation: `${venueId}/${secondaryAreaChairName}/-/Assignment`,
```
melisabok, line 374: is it possible to pass the whole invitation id in the property of the component?

melisabok:
@xkopenreview do you have a screenshot for this?

I want to check if there is a "Copy Email" link in the secondary area chair info.

components/webfield/AreaChairConsole.js
melisabok: @xkopenreview I left a comment here

melisabok:
I don't understand this change. Why an AC has assigned a Secondary AC?

melisabok (approved):

enrubio:
So we create this invitation: `{venue_id}/{secondary_area_chairs_name}/-/Assignment`. Should the head & tail be this?:
```
"param": {
      "type": "profile",
      "options": {
        "group": "venue_id/Area_Chairs"
      }
    }
```
And each triplet member has an assignment edge to each other?

enrubio:
> suggest to prepare the workflow where this is needed

What do you mean by this? Do you mean making changes in openreview-py so it creates this invitation? Maybe if `secondary_area_chairs` is enabled in the request form we can create it.

melisabok:
I don't think all the venues with secondary area chairs will have this assignment. I think this is a feature related to CVPR/ICCV venues where they have triplets. 

My suggestion is to create the assignment invitation in the test where we are testing triplets and go from there.

enrubio:
Let me check again

enrubio:
Thank you for working on this Xukun. Discussed this with Melisa and since we don't currently represent secondary ACs as edges we can close this. We think we should use groups instead.

PR #2229, opened 2025-02-10
Fix/ Assignments - read preferredEmailInvitationId from domain

xkopenreview:
edge browser required preferredEmailInvitationId query param to show copy email link in profile entitiy

this pr should read domain object for preferredEmailInvitationId
so that the param can be added to edge browser url if it exists


melisabok (approved):
Thanks!

PR #2230, opened 2025-02-10
Fix/ Messages - loading status when no message is found

xkopenreview:
related to #2216 where the loading status is broken when there's no message

this pr should fix the loading status


melisabok (approved):

emily-grabowski (approved):
I do get a flicker of the loading message when I type in a non-existent email, but perhaps that is expected.

emily-grabowski:
Makes sense! Definitely better than before

PR #2234, opened 2025-02-20
Fix/ AC console tasks - add extraRoleNames config

xkopenreview:
this pr should add a config (string array)
>extraRoleNames

to ac console

the purpose is to show invitations where invitee is secondary area chairs


melisabok:
Do you mind sharing a screenshot of the change?

where this change is being made, this is modifying the task tab?

enrubio (approved):

melisabok:
I see, I would expect to see one task tab per role so you know which tasks correspond to each role. What do you think?

enrubio:
Oooh, I really like that idea. I think that will help them quickly find which papers they need to confirm. @xkopenreview do you think we can do that?

melisabok (approved):

enrubio (approved):
It looks great! Thanks Xukun.

Screenshot 2025-03-21 at 3 44 28 PM

PR #2237, opened 2025-02-28
Fix/ PC/SAC Console - Add latest reply column in paper status tab

xkopenreview:
this pr should add in PC and SAC console paper status tab a new column (when displayReplyInvitations is specified) to show latest reply of the invitation and content fields specified.

sample config
```javascript
displayReplyInvitations: [
  {
    id: "acmmm.org/ACMMM/2024/Conference/Submission{number}/-/Official_Review",
    fields: ["summary", "limitations"],
  },
  {
    id: "acmmm.org/ACMMM/2024/Conference/Submission{number}/-/Public_Comment",
    fields: ["strengths", "suitability"],
  },
]
```

### DIFF
--- a/components/webfield/AreaChairConsole.js
+++ b/components/webfield/AreaChairConsole.js
@@ -25,6 +25,7 @@ import {
   pluralizeString,
   getSingularRoleName,
   getRoleHashFragment,
+  buildNoteTitle,
 } from '../../lib/utils'
 import { referrerLink, venueHomepageLink } from '../../lib/banner-links'
 import AreaChairConsoleMenuBar from './AreaChairConsoleMenuBar'
@@ -571,6 +572,7 @@ const AreaChairConsole = ({ appContext }) => {
                   value,
                 }
               }),
+              signature: latestReply?.signatures?.[0],
             }
           }),
         }
--- a/components/webfield/AreaChairConsole.js
+++ b/components/webfield/AreaChairConsole.js
@@ -2,13 +2,14 @@
 
 import { useContext, useEffect, useState } from 'react'
 import { useRouter } from 'next/router'
+import { orderBy } from 'lodash'
 import WebFieldContext from '../WebFieldContext'
 import BasicHeader from './BasicHeader'
 import { Tab, TabList, TabPanel, TabPanels, Tabs } from '../Tabs'
 import Table from '../Table'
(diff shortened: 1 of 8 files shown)


enrubio:
What do you think about limiting the number of characters/lines displayed from a note? Is that possible? For example:

latest-reply

melisabok:
Do you mean to shop only the first two lines and then they can click on the title to read the whole comment in the forum page?

enrubio:
Yes exactly, what do you think?

melisabok:
I'm not sure what the users would wan to see. The latest comment or the latest activity.

I understood from Yoshitomo that they want to see the discussion between AC/SACs in the console without having to go to each forum. This column is not intended to be used to render a review, it is more about comments, the questions is how long this comments would be.

enrubio:
Yes, I was just suggesting to limit the number of characters displayed in the console since it could be a very long comment.

Another thing:

> I understood from Yoshitomo that they want to see the discussion between AC/SACs in the console without having to go to each forum.

Right now we show the latest comment reply regardless of who it's from, so they'll see reviewer comments as well. For their purpose, I don't think they want to see those right? Since they want help with tracking emergency reviewer assignment progress. So should we do any kind of filtering on the comments that get displayed?

This is based on Yoshitomo's comment in the doc:

> What about showing the latest comment only visible to SAC (PC) in SAC (PC) console instead of using chat feature? That way, I think SAC (PC) can see what the latest comment from the AC are for SAC (PC) e.g., looking for emergency reviewers, contacted reviewers, etc

Seems like he's asking for filtered comments.

melisabok:
I was planning to use a specific comment or chat invitation for discussion between AC/SACs

enrubio:
Ohhh okay, that makes sense. Then I think it looks good in all consoles. So do you think it's necessary to truncate the text? If not, I can approve it as is.

melisabok:
- Let's not truncate it for now so the user can read the whole comment from the console.
- Let's not filter the replies by readers to not add more complexity. We can add it later if needed.

enrubio:
Ok! That sounds good, thank you for clarifying.

enrubio (approved):
Checked PC/SAC/AC consoles and the changes look good! Thank you Xukun.

melisabok (approved):

Issue #2238, opened 2025-03-03
Console Tasks: Mark tasks as complete if the note contains all mandatory fields


enrubio:
Instead of marking a task as complete if a note has been submitted with that invitation, we can check if the note has all required fields filled out.

PCs can change forms and required fields while a stage is going on, so this would help the committee see that they have unsubmitted fields.

melisabok:
Emilia, is there deadline May 23 accurate? I would like to use a deadline that is previous to when we need it so we can have enough time to deploy it.

enrubio:
They'll need this for the Meta Reviews and it seems the Meta Review Stage will start May 28. I updated the deadline to 2 weeks before then

PR #2249, opened 2025-03-10
Fix/ Console Tasks: Mark tasks as complete only if the note contains all mandatory fields

xkopenreview:
this pr should update the task completion logic in console tasks and tasks page to
besides checking for existence of reply note/reply edit
it should also check that the reply note/reply edit should contain all non-optional fields of the invitation to be considered as completed

so that invitation change can correctly display status of notes submitted before the change

Fixes https://github.com/openreview/openreview-web/issues/2238

### DIFF
--- a/components/profile/ProfilePreviewModal.js
+++ b/components/profile/ProfilePreviewModal.js
@@ -19,7 +19,6 @@ const ProfilePreviewModal = ({
   sortFn,
   showNextProfile,
   acceptUser,
-  blockUser,
   setProfileToReject,
   rejectUser,
 }) => {
@@ -31,11 +30,8 @@ const ProfilePreviewModal = ({
   const [rejectionReasons, setRejectReasons] = useState([])
   const needsModeration = profileToPreview?.state === 'Needs Moderation'
 
-  const updateMessageForPastRejectProfile = () => {
-    setRejectionMessage(
-      (p) =>
-        `Submitting invalid info is a violation of OpenReview's Terms and Conditions (https://openreview.net/legal/terms) which may result in terminating your access to the system.\n\n${p}`
-    )
+  const updateMessageForPastRejectProfile = (messageToAdd) => {
+    setRejectionMessage((p) => `${messageToAdd}\n\n${p}`)
   }
 
   const loadPublications = async () => {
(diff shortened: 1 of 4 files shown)


enrubio:
About Xukun's comment here: https://github.com/openreview/openreview-web/issues/2238#issuecomment-2714820598

I agree that users will probably be confused since we're changing the way the task status is considered completed. I'm worried we'll get many emails claiming it's a bug. Is it possible to convey this to the user somehow?

This isn't a perfect example and I don't think we should necessarily do this, but is there a simple change like this we could make?:
ac-console-tasks

I'm not sure if we should make a distinction between a task with a submitted note that needs to be updated vs a task with no note (since both will show as incomplete). Does anyone have any thoughts or suggestions? @melisabok @xkopenreview @celestemartinez

melisabok:
Can we show the task as pending but describe there are fields missing?

enrubio:
This is what the changes look like:

console-tasks


Minor suggestion to make the font size of "Update is required" to be the same as the due date font size. I'm hoping it reads better, because otherwise when you read from left to right we go from big font, to small, to big again.

console-task-suggestion

enrubio:
This is what the changes look like now:
Screenshot 2025-04-28 at 4 39 15 PM

I think I liked it better when it was bolded to bring your attention to it:
Screenshot 2025-04-28 at 4 39 25 PM

Otherwise I think this looks good. Maybe in the future we can have a status column since tasks can now have 3 statuses: unsubmitted, submitted, and update needed

haroldrubio:
# Generated Proposed mockup

*(it generated an image that's cropped)*
ChatGPT Image Apr 28, 2025 at 10_27_33 AM

# Generated Feedback
| Issue | Why it’s a problem | Minimal-change fix |
|-------|-------------------|--------------------|
| Visual competition with the due-date pill (both are red-ish, same weight). | The two elements fight for attention and make the row feel noisy. | Give the status its own tint (e.g., orange) and put the text inside a pill so it’s visually distinct but less alarming. |
| Alignment / spacing – “Update is required” sits right after the due-date pill, creating an awkward ragged edge and variable whitespace. | Harder to scan; rows feel uneven. | Reserve a narrow right-hand column for status pills; keep equal inset on every row. |
| Repetition fatigue – six identical strings dominate the screen. | Users quickly tune out repeated copy; the page looks busier than it is. | Shorten to something like “Requires update” (two words) or even an icon + tooltip. |
| Color-blind accessibility – relying on red alone penalises users with deuteranopia/protanopia. | Status may be missed. | Pair the colour with the pill outline and/or an icon (⚠︎) so shape + text convey meaning. |
| Tone – dark red implies an error, not a routine nudge. | Creates unnecessary urgency. | Use a warning/attention colour (amber) unless the task is truly overdue, then switch to red. |

melisabok:
I like @haroldrubio's screenshot but I can not see the full text, can you share the part that shows the update is required?

melisabok (approved):

PR #2259, opened 2025-03-31
Fix/ Invitation Editor - Add ddate

xkopenreview:
this pr should allow ddate to be added/removed in invitation editor


components/group/GroupRelatedInvitations.js
```diff
@@ -28,6 +28,7 @@ const GroupRelatedInvitations = ({ group, accessToken }) => {
       : {
           prefix: groupId.includes(submissionName) ? `${groupId}/.*` : `${groupId}/-/.*`,
           expired: true,
+          trash: true,
```
melisabok, line 31: @carlosmondra said you can not get all the expired and trashed invitations in the same call, is that right?

melisabok (approved):
It works great!

carlosmondra (approved):

PR #2262, opened 2025-04-04
Fix/ Test setup - make sure the order of cdate <= duedate <= expdate

xkopenreview:
similar change to github.com/openreview/openreview-py/pull/2549
otherwise revision process would fail and post submission date process will get stuck


melisabok (approved):
Thanks!

PR #2263, opened 2025-04-07
Fix/ Edge browser - replace dot in filter query

xkopenreview:
filter param in edge browser is in the form of 
browseInvitationId = value

where browse edges are added to row data as row.filterProperties.browseInvitationId

however when browseInvitationId contains dot (for example thecvf.com/...), the dot will be interpreted as a level of field in row and the filtering logic will compare the value of row[filterProperties][thecvf][com/...] and fail

this pr should replace all dot in filtering invitation to underscore

### DIFF
--- a/components/browser/Column.js
+++ b/components/browser/Column.js
@@ -616,9 +616,10 @@ export default function Column(props) {
         ['!=', '>=', '<=', '>', '<', '==', '='],
         editAndBrowserInvitationsUnique.reduce(
           (prev, curr) => {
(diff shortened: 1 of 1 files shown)


melisabok (approved):

enrubio (approved):
This works great, thank you! Tested with this PR: https://github.com/openreview/openreview-py/pull/2559 and emergency reviewers are correctly being filtered out. This allows ACs to also sort them by affinity score.

Ex:
Screenshot 2025-04-07 at 3 22 11 PM

PR #2278, opened 2025-04-28
Fix/ Note Editor: show a warning message when the readers are a subset of the parent.readers

xkopenreview:
show warning message when note reader might be incomplete

### DIFF
--- a/app/Footer.js
+++ b/app/Footer.js
@@ -116,7 +116,7 @@ export default function Footer() {
         </div>
       </footer>
 
-      <footer className="sponsor">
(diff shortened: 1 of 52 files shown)


melisabok:
Screenshot 2025-05-01 at 2 52 50 PM

I shouldn't see the warning in this case, the signature is member of the action editor group, is it possible to verify that?

melisabok:
When the comment is public, I get the warning and I shouldn't:

Screenshot 2025-05-07 at 9 25 11 AM

PR #2279, opened 2025-04-29
Fix/ Anon group readers when reviewer is unassigned

xkopenreview:
this pr should add support of inGroup type in reply readers

members of the specified inGroup group will be added to options


melisabok:
Yay! Thanks Xukun, let me add it to the openreview-py comment invitations and let's start using it!

melisabok:
It works great! Thank you Xukun. I notice that we are using prefix signatures.items. Do you think it can be supported too?

melisabok:
I think we need to keep using prefix in the signatures because the UI needs to query groups by signatory

melisabok (approved):

PR #2294, opened 2025-05-12
Fix/ Parse whole string when there's no colon

xkopenreview:
this pr should update the parse number util function to handle colon typo in invitation


melisabok (approved):
Thanks!

PR #2297, opened 2025-05-14
Fix/ PC Console - cache data

xkopenreview:
this pr should add a new config 
>useCache

to pc console config
when set to true, it should cache in browser indexed db most pc console data so that when user does not need latest data it can load faster.

when user access pc console, cache older than 24 hours are removed.


melisabok:
What about 1, 4 and 5, are you working on it?

carlosmondra (approved):

PR #2300, opened 2025-05-14
Fix/ Group members - replace append action with add

xkopenreview:
group member is a set so the action should be
>add

instead of 
>append

which is used for array


melisabok:
Closing this one, it is being fixed here: https://github.com/openreview/openreview-web/pull/2350

PR #2303, opened 2025-05-15
Fix/ Profile Moderation: create tag to organize profile moderation

xkopenreview:
related to github.com/openreview/openreview-py/pull/2585
to support tagging profile in moderation


melisabok:
Can the tag be next to widget to post the tag?

Screenshot 2025-05-28 at 3 24 15 PM

melisabok:
could try showing the tags in the profile page too?

melisabok:
I would like the moderator use the tags as much as they need. These tags should be visible to the support team only.

I believe we profiles are merged, the tags as merged too, right?

melisabok:
Closing, no need it now.

PR #2308, opened 2025-05-21
Nextjs 15 upgrade 

xkopenreview:
same as #2245 
#2245 is closed due to branch rename to remove /

### DIFF
--- a/components/NoteEditor.js
+++ b/components/NoteEditor.js
@@ -241,8 +241,12 @@ export const getNoteReaderValues = async (
       invitation.edit.note.readers?.param?.items?.map(async (p) => {
         if (p.value) return p.value
         if (p.inGroup) {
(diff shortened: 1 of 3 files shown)


app/assignments/Assignments.js
```diff
@@ -82,28 +64,18 @@ export default function Assignments({ accessToken }) {
         <LoadingSpinner />
       </CommonLayout>
     )
+  if (configInvitation.apiVersion !== 2)
+    return <ErrorDisplay message={'Old version of assignments is not supported'} />
```
melisabok, line 68: Please use this message: "This venue was hosted using a deprecated API version, and the assignments are not available through this page. Please use our contact form if you need to access the assignment data."

melisabok (approved):

PR #2322, opened 2025-06-09
Fix/ Profile Moderation: create tag to organize profile moderation (nextjs 15)

xkopenreview:
changes of #2303 based on nextjs 15


melisabok:
> in the test i saw Reviewers_Review_Count tags being returned which does not have a label

If no label is provided the weight should be shown. These tags are intended to be shown in the profile page too. The tag label and weight has the same behavior than edge: both are optional but at least one of them must be present.

melisabok (approved):

PR #2288, opened 2025-05-06
Fix/ Remove the info email from the Contact page

xkopenreview:
this pr should remove feedback modal and move the function to contact page
and make some adjustment to the feedback form fields

PR #2324, opened 2025-06-10
Fix/ Remove the info email from the Contact page (nextjs15)

xkopenreview:
changes of #2288 based on nextjs 15  (v1.14.0)

### DIFF
--- a/app/contact/page.js
+++ b/app/contact/page.js
@@ -25,7 +25,7 @@ export default function Page() {
   const { turnstileToken, turnstileContainerRef } = useTurnstileToken('feedback')
   const { user, accessToken, isRefreshing } = useUser()
 
-  const profileSubject = 'My OpenReview profile'
+  const profileSubject = 'I have a question about my existing OpenReview profile'
   const submissionSubject = 'A conference I submitted to'
   const organizationSubject = 'A conference I organized'
   const committeeSubject = 'I am a reviewer or committee member'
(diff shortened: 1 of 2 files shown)


melisabok:
I think offering only these 4 options is too limited:

Screenshot 2025-06-30 at 11 44 18 AM

Could we include cases like:
- the user is a reviewer or member of the committee
- the user is trying to create their profile
- the user is trying to access to a publication

melisabok:
If I have a question about my profile, it is asking the venue id, can this be optional? some users might to register to openreview to read papers. 

Screenshot 2025-06-30 at 1 03 10 PM

melisabok (approved):

enrubio:
The form looks great. I see we have 2 dropdown options relating to profiles: `My OpenReview Profile` and `I am trying to create my profile`. I feel we can have just one profile related option, maybe the first one. What do you think?

melisabok:
Some users are already registered in OpenReview and they may have issues with their existing profile so I would keep both.

enrubio:
Are we not allowing guests to contact us through the feedback form? I see: `Error: signature must match pattern "^~.+\d+$|.+@.+\..+|^(guest)$"`.

emily-grabowski:
I suggest renaming the following option:
- 'My OpenReview profile' --> I have a question about my existing OpenReview profile
so that users know which one to choose of the profile options

My other suggestion would be to ask what someone's role is in a venue (Author, Reviewer, Area Chair, etc), but I would understand if this adds too much complexity

enrubio (approved):

PR #2357, opened 2025-07-08
Fix/ Enhance token handling - Add rate limit and expired token error handling

xkopenreview:
this pr should
- handle rate limit error from /refreshToken call
- handle the case that a cached expired access token being returned from /refreshToken
- add some test to server auth and client auth

### DIFF
--- a/lib/api-client.js
+++ b/lib/api-client.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-use-before-define */
 import { stringify } from 'query-string'
-import { getTokenPayload } from './clientAuth'
+import { shouldTryRefreshToken, getTokenPayload } from './clientAuth'
 
 const formatError = (err) => {
   if (process.env.SERVER_ENV !== 'production') {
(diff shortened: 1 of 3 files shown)


lib/api-client.js
```diff
@@ -37,7 +38,16 @@ const checkStatus = async (response, originalRequest) => {
       },
     })
     const data = await refreshResponse.json()
-    if (data.name === 'TokenExpiredError' || data.name === 'MissingTokenError') {
+    if (
```
melisabok, line 41: why do you need to check for only these 3 errors instead of any error?

lib/api-client.js
```diff
@@ -38,11 +38,7 @@ const checkStatus = async (response, originalRequest) => {
       },
     })
     const data = await refreshResponse.json()
-    if (
-      data.name === 'TokenExpiredError' ||
-      data.name === 'MissingTokenError' ||
-      data.name === 'RateLimitError'
-    ) {
+    if (!data.token) {
       throw new Error('Token has expired, please log out and log in again.')
```
melisabok, line 42: can you add a log here so we can see the type of errors we are getting?

melisabok (approved):

PR #2359, opened 2025-07-09
Fix/ AC/SAC/PC Console - support default message body when email reviewers

xkopenreview:
requestd by NeurIPS 25
When AC/SAC message reviewers from console, some reviewers think the email is from PC and contacted PC. 

this pr should add a defaultEmailBody config so that the console can specify that this email is from AC/SAC

```
defaultEmailBody:'\n\n\nThis email is from AC.\n\nPlease contact AC directly'
``` 

should render to:
```


Sent by SAC.
Please Contact SAC directly

Thanks
```


melisabok:
I don't think this is the right solution, there is an issue that requests the user to select the replyTo of the message. The replyTo must be either the user that is sending the reminder or the PC contact info.

PR #2367, opened 2025-07-14
Fix/ Profile page - publication published/modified date discrepancy in client rendering

xkopenreview:
Some notes in profile publications page may be rendered to different dates in server and client
and cause hydration error

this pr should render the publication dates in profile page only in the browser


melisabok:
Can you explain how to reproduce this error?

melisabok (approved):

PR #2369, opened 2025-07-15
Fix/ Invitation editor: curly braces are removed when the id has several ones

xkopenreview:
this pr should disable TeX rendering in the code editor
so that mathjax does not mess with $ in code editor


melisabok (approved):

melisabok:
Thanks, I would like to include this in the next release.

PR #2377, opened 2025-07-29
Fix/ Profile publication not shown

xkopenreview:
this pr should show loading when loadPublication fail

this is to check the issue that publication and coauthor show empty in profile page

### DIFF
--- a/app/profile/Profile.js
+++ b/app/profile/Profile.js
@@ -45,7 +45,12 @@ export default async function Profile({ profile, publicProfile }) {
       limit: 1000,
     }
     try {
-      apiRes = await api.getCombined('/notes', queryParam, null, { accessToken: token })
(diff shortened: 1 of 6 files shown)


app/profile/Profile.js
```diff
@@ -46,6 +46,12 @@ export default async function Profile({ profile, publicProfile }) {
     }
     try {
       apiRes = await api.getCombined('/notes', queryParam, null, { accessToken: token })
+      if (apiRes.notes) {
+        publications = apiRes.notes
+        // eslint-disable-next-line prefer-destructuring
+        count = apiRes.count
```
carlosmondra, line 52: Don't you need to pass count=true in the queryParam?

carlosmondra: It's offset + limit that is equivalent to count=true

PR #2383, opened 2025-08-04
Feat/ ProfileSearch - add drag sort function

xkopenreview:
this pr should add drag sort function when there are >5 authors


melisabok:
it looks good to me but I wonder why we enable this for > 5 authors instead of doing for any author list length.

racheljsmart (approved):

PR #2389, opened 2025-08-07
Fix/ Profile Search - multiple authors with same email should not be allowed

xkopenreview:
this pr should check for custom author email in already added authors and disable the add custom author button if the email has been added

not an issue for fields other than authorid as custom author is not allowed in other fields


melisabok:
Shouldn't this be an API validation instead of a UI fix?

melisabok (approved):

PR #2391, opened 2025-08-08
Fix/ Improve accessibility

xkopenreview:
this pr should fix most accessibility errors

most of the errors are not enough contrast so the background/text color and opacity are changed


melisabok:
@celestemartinez could you review the new UI style?

PR #2409, opened 2025-08-25
Fix/ Profile editor - handle relations with no readers

xkopenreview:
some old profiles have relations with no readers which will lead to error when user edit relations in profile edit

this pr should add a fallback value for relation readers


enrubio (approved):
Changes look good

PR #2425, opened 2025-09-02
Fix/ PC Console overview - move invited groups logic to main loading

xkopenreview:
this pr should move calls to get invited groups to pc console main data loading
so that it can be cached

user may need to reload data if having old cache which does not include invited count


melisabok:
It works great

melisabok (approved):

PR #2444, opened 2025-09-12
Fix/ Reviewer Console - add doc for config

xkopenreview:
this pr should add doc for variables used in reviewer console config
this pr should also fix the issue with custom load comparison and fix missing config logic for recruitmentInvitationId
it used to be for v2 only and is moved to the list of missing configs as v1 logic is removed.

### DIFF
--- a/components/webfield/ReviewerConsole.js
+++ b/components/webfield/ReviewerConsole.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 /* globals typesetMathJax,promptError: false */
 import { useContext, useEffect, useState } from 'react'
 import { useSearchParams } from 'next/navigation'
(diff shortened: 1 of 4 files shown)


melisabok:
What is `reviewDisplayFields` for?

components/webfield/ReviewerConsole.js
```diff
@@ -367,6 +368,163 @@ const ReviewerConsoleTabs = ({
...
+ * @name ReviewerConsoleConfig.customMaxPapersInvitationId
+ * @description Related to recruitmentInvitationId and reviewLoad. The invitation to get custom load edge. If edge exist the weight is used as custom load otherwise it will load the recruitment note to read the reduced_load field.
+ * @type {string}
+ * @default no default value
+ * @example
+ * { "customMaxPapersInvitationId": "ICLR.cc/202X/Conference/Reviewers/-/Custom_Max_Papers" }
+ */
+
+/**
+ * @name ReviewerConsoleConfig.reviewLoad
```
melisabok, line 501: why do we have this setting?

melisabok: ohhh it is hardcoded.

components/webfield/ReviewerConsole.js
```diff
@@ -367,6 +368,163 @@ const ReviewerConsoleTabs = ({
...
+ * @description Related to recruitmentInvitationId and customMaxPapersInvitationId. The default value to display in header when there's no custom load edge or recruitment note
+ * @type {string|number}
+ * @default no default value
+ * @example
+ * { "reviewLoad": "ICLR.cc/202X/Conference/Reviewers/-/Custom_Max_Papers" }
+ */
+
+/**
+ * @name ReviewerConsoleConfig.hasPaperRanking
```
melisabok, line 510: I don't think we support paper ranking in the workflow anymore so please remove it

PR #2447, opened 2025-09-16
Fix/ Tasks page - split data loading

xkopenreview:
currently the tasks page loads all invitations with all reply data

this pr should change it to
- load domain only, still 3 calls to get note,tag,edge invitations
- load full invitation data when user click load tasks


melisabok:
I do not see any network call being done after click the arrow to expand the tasks:

Screenshot 2025-09-18 at 1 19 56 PM

melisabok:
I'm testing this as test@mail.com and running the openreview-py tests

PR #2449, opened 2025-09-17
Fix/ Forum - show link to to other versions

xkopenreview:
this pr should add links to other versions of the forum note below forum meta
when it's available by paperhash


melisabok:
let's keep it as it is then.

PR #2477, opened 2025-10-10
do not use offset for invitations calls in API 1

melisabok:


client/webfield.js
```diff
@@ -134,6 +134,7 @@ module.exports = (function () {
         resultsKey = 'profiles'
       } else if (url.indexOf('invitations') !== -1) {
         resultsKey = 'invitations'
+        offset = undefined
```
xkopenreview: looks like this is not enough
tags param is also disabled.
@melisabok please take a look at https://github.com/openreview/openreview-web/pull/2483/commits/5fb6c49a9969e4e5451868ae225b82026bacae35

PR #2490, opened 2025-10-16
Fix/ Forum page - Destroy note editor when active invitation change

xkopenreview:
this pr should fix the following issue:
when there are multiple reply invitations with similar definition in the same forum page, the lazy loaded widget component for a field (such as radio button) may be reused by react and cause the widget to render stale content if user switch between these reply invitations.

by destroy the note editor when active invitation for reply invitation is changed


enrubio:
thank you!

PR #2493, opened 2025-10-17
Feat/ Moderation - replace Message with Past State in profile preview modal

xkopenreview:
this pr should remove messages section from profile preview modal
and replace it with past states which shows more info
it will show the content of rejection email if one is sent within 5 seconds of the state change


melisabok:
Nice!

PR #2495, opened 2025-10-20
Fix/Orcid - Check credit name

xkopenreview:
currently orcid import is checking firstname+lastname and other names to verify ownership of orcid url
some profiles also have a credit-name
so this pr adds it in the check

picture below illustrates the different names in orcid page
image

### DIFF
--- a/.testcaferc.json
+++ b/.testcaferc.json
@@ -1,5 +1,6 @@
 {
   "browsers": "chrome",
   "videoPath": "videos",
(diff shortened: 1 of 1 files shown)


xkopenreview:
> is there any change needed in the pre-process validation for the authorship claim?

no

this is the orcid url validation
the process/authorship cliam is between name in a work and profile name

melisabok (approved):

PR #2498, opened 2025-10-20
Fix/articles news

melisabok:
- add referrer param to go back to home page
- support cdate and mdate to be editable through the editor
- skip colon if fieldName is empty string

### DIFF
--- a/.testcaferc.json
+++ b/.testcaferc.json
@@ -3,5 +3,6 @@
   "videoPath": "videos",
   "stopOnFirstFail": true,
   "pageRequestTimeout": 30000,
+  "hostname": "localhost",
   "retryTestPages": true
 }


app/(Home)/News.js
```diff
@@ -42,9 +42,9 @@ export default function News({ news }) {
       />
       <button type="button" className="btn-link">
         <a
-          href={`/group?id=${process.env.SUPER_USER}/News`}
+          href={`/group?id=${process.env.SUPER_USER}/News&referrer=[Homepage](/)`}
           target="_blank"
-          rel="nofollow noreferrer"
```
xkopenreview: i think this should not be removed
as target is still _blank

melisabok: done

client/view-v2.js
```diff
@@ -2418,6 +2424,9 @@ module.exports = (function () {
           case 'license':
             editNote[p] = formContent.noteLicenseValue ?? edit.note[p]
             break
+          case 'cdate':
+            editNote[p] = formContent.noteCDateValue ?? edit.note[p]
+            break
           case 'pdate':
```
xkopenreview: constructUpdatedEdit does not have mdate is it intentional?

melisabok: thanks for the catch, added

xkopenreview:
[6eb2025](https://github.com/openreview/openreview-web/pull/2498/commits/6eb20252e5a54199794a9a618016d912afb7e8b4) should fix the issue that

date can't be selected when editing edit c/m/pdate in revisions page
because date picker panel is rendered outside the modal
so clicking on the date picker reset it
6eb2025 solves the issue by rendering date picker panel inside modal window

Issue #2501, opened 2025-10-22
Health check: build a new endpoint to check the openreview UI is up and can connect to the APIs


melisabok:
Use this health check by the instance groups to delete unhealthy instances. 

The endpoint should be called: openreview.net/healthcheck

and it can try to render a page, connect to API 1, connect to API 2

xkopenreview:
web has a /version route 
if it returns if means web is running

web's only dependency is api
which can be verified by calling health check of api?

melisabok:
I think it is important to check if the instance has access to the API that is different to the API health check.

I would still add the new route thought to be consistent with the other services.

PR #2502, opened 2025-10-22
Fix/ Add web health check route

xkopenreview:
this pr should add a route to perform health check of

- web app itself
- connectivity from web app to api

to be called by google cloud health check

### DIFF
--- a/app/healthcheck/route.js
+++ b/app/healthcheck/route.js
@@ -15,5 +15,5 @@ export async function GET() {
       error,
     })
   }
-  return Response.json({ healthy: false })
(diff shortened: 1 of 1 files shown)


xkopenreview:
not to be confused with  existing web health check
which is used by load balancer to perform app-wise check

this router is for instance-wise health check

melisabok:
Fixes https://github.com/openreview/openreview-web/issues/2501

melisabok:
> not to be confused with existing web health check
which is used by load balancer to perform app-wise check

What is the existing health check? calling to openreview.net?

PR #2509, opened 2025-10-27
Fix/ Integration tests update

xkopenreview:
this pr should run 2 instances of testcafe so that tests can run faster

concurrency option of testcafe does not work very well so only added for setup

### DIFF
--- a/app/(Home)/Home.module.scss
+++ b/app/(Home)/Home.module.scss
@@ -112,3 +112,18 @@
     min-width: fit-content;
   }
 }
+
+.closeNewsBtn {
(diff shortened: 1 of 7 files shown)


xkopenreview:
about 1min faster
but logs are less clear, screencapture seems broken too
maybe it's not worth it

melisabok (approved):

carlosmondra:
> about 1min faster but logs are less clear, screencapture seems broken too maybe it's not worth it

Yeah, it doesn't seem to be worth it

carlosmondra (approved):

PR #2516, opened 2025-10-28
Fix/ Add Tags viewer

xkopenreview:
this pr should add a new component 
TagsViewer

which shows profiles of specified tag and whether the user is member of any groups in specified domain


melisabok:
How do I access to this page? is this a new React component?

melisabok (approved):
Thanks Xukun!

PR #2523, opened 2025-11-03
Fix/ Profile search - add institution

xkopenreview:
this pr should add a new note editor widget to support new schema of authors with institutions or without institutions( have only 2 properties username and fullname)

related to 
https://github.com/openreview/openreview-py/pull/2741 and
https://github.com/openreview/openreview-api/pull/874


xkopenreview:
@melisabok is it a valid use case for a venue to use object authors but do not want institutions?
or it should use author+authorids

melisabok:
> is it a valid use case for a venue to use object authors but do not want institutions?

Yes, it should be valid. I'm planning to use an object authors for the DBLP, ORCID import.

My believe is that once this is working all new venues should start using objects to represent authors, with or without institutions.

melisabok (approved):

celestemartinez (approved):

carlosmondra (approved):

PR #2525, opened 2025-11-06
Fix/ Profile - replace tags with service roles

xkopenreview:
this pr should replace profile tags with service roles
which shows only role related tags


melisabok:
> @melisabok do we want to keep the su only logic

yes, until I can create all the tags for all the venues, then we release it to the public

PR #2537, opened 2025-11-19
Fix/ WorkflowInvitations - changes from all hands

xkopenreview:
this pr should:
1. Rename link “details” to “logs”
2. Show the “logs” link even when the process is running
3. changing group loading from parent then filter out submission groups to load specified groups and sub groups
4. add filterStaticForum param in call to get invitations


celestemartinez (approved):
Looks good!

PR #2540, opened 2025-11-20
Fix/ Move “active consoles” to the user menu

xkopenreview:
this pr should move active consoles to nav bar

size of search input is changed to fit the active consoles


melisabok (approved):

xkopenreview:
briefly discussed about this
rationale of moving active consoles to nav bar is to make space for some new components such as stats

original idea of showing active consoles as single dropdown in nav bar is not practical, especially on small screens
tried moving active console to a separate page with role/venue filtering and have a link to the new page in nav bar
there are two issues with this approach:
- it makes the active console more difficult to access
- there's not much value add moving active consoles to a separate page. the filtering is probably useless for most users 

@carlosmondra proposed another approach:
keep the homepage same as it is for guest
when user login, replace active venues with active consoles
as active venues is probably the least useful info 
and search venues results will show tag if the venue is active

i think it is better then moving active consoles to nav bar or new page. @melisabok please comment

melisabok:
I like Carlos's idea, let's experiment that.

xkopenreview:
#2766 is already merged so close

PR #2543, opened 2025-11-24
Fix/ SAC Console - load notes by number

xkopenreview:
this pr should change notes loading logic from
- load all ->filter those assigned
to
- load only the assigned

to avoid exceeding url length limit, the assigned notes are broken to batches of 50 number each


xkopenreview:
if icml experiment use rating name config then this pr is complete
otherwise also need to support display and sorting of new fields

melisabok (approved):

PR #2546, opened 2025-11-25
Feature/ Replace all venues with search venues

xkopenreview:
this pr should replace all venues in homepage with venue search
requires https://github.com/openreview/openreview-api/pull/878


app/(Home)/AllVenuesWithSearch.js
```diff
@@ -0,0 +1,196 @@
...
+const wordSegmenter = new Intl.Segmenter(undefined, { granularity: 'word' })
+
+const tokenizeTerm = (term) => {
+  if (!term) return []
+  return Array.from(wordSegmenter.segment(term))
+    .filter((s) => s.isWordLike)
+    .map((s) => s.segment)
+}
+
+const truncateAroundMatch = (label, tokenizedTerm) => {
```
xkopenreview, line 22: truncateAroundMatch and highlightMatch is kind of similar to truncate used in autocomplete search in nav bar but support multiple terms

autocomplete search in nav bar should be changed to use this function when converted to antd

xkopenreview:
[dc3a3df](https://github.com/openreview/openreview-web/pull/2546/commits/dc3a3dfa23461382828e7e78eb195bd654733e05) is showing the es down message in the dropdown because homepage has the donate banner as the default banner

xkopenreview:
[5d5fb9b](https://github.com/openreview/openreview-web/pull/2546/commits/5d5fb9b585d01a1839577de55bed994cdacf9420) should:
1. pass full search term (trimmed) to api
2. show active and open for submission tag
3. pass select to reduce data returned by api

immediateSearchTerm may not always trigger token change when it does not contain word
this is intentional to avoid meaningless extra api call
for example when user search "cvpr", it will make an api call
when user append ...... to make the search input to show "cvpr......" will not trigger api call

xkopenreview:
looks like having bot only links by hiding it from human users is anti-pattern
correct fix is to have footer pointing to /all-venues
and add /venues to sitemap.xml so that bot goes to /venues directly

melisabok:
looks good! some suggestions:

- Given that we are going to show two search textboxes in the homepage, let's make clear what they are for:

Header placeholder can say: "Search articles, authors and reviews..."
Venues search place holder can say: "Search venues by name..."

- The all-venues page is more like a directory of venues because we do not show a flat list of all the venues, we only show the root group. So I would change it to be /venues-directory,  /venues-browse, /venues/browse. An change the string "View All Venues..." with "Browse all venues"

- The bot related pages can be renamed as /sitemap/venues, /sitemap/venue?id=, /sitemap/submissions?venue=

carlosmondra (approved):
Works!

xkopenreview:
[b99df13](https://github.com/openreview/openreview-web/pull/2546/commits/b99df13c8b5648a1a9ee27824ebb3cac9c5826b3) should
- change view all venues link to browse all venues
- update placeholder text in the search in nav bar

discussed about the route changes
decided to do that in another pr because it involves converting footer to antd, load balancer update and sitemap.xml setup which are not closely related to changes in this pr

melisabok (approved):

cagostino (approved):
changes look good, homepage all-venues list replacement with /venues/search, a-z list moved to /all-venues with letter grouping. /venue and /venues moved under bot route group

PR #2549, opened 2025-11-26
Feature/ Add custom stage invitation fields to sort options

xkopenreview:
this pr should add customStageInvitaitons config displayField and extraDisplayFields to sort options


enrubio:
I see. The naming isn't consistent, sometimes there are fields with the same name as the invitation and sometimes not. I guess we can keep it for now and change it later if needed.

PR #2554, opened 2025-12-01
Fix/ Signup page - new sign up only

xkopenreview:
this pr should:
1. allow new sign up only in sign up page
2. revert #2552 
3. clean up unused get profile by dblp function
4. update healthcheck endpoint
it depends on https://github.com/openreview/openreview-api/pull/911

### DIFF
--- a/components/webfield/RecruitmentForm.js
+++ b/components/webfield/RecruitmentForm.js
@@ -14,6 +14,7 @@ import { translateInvitationMessage } from '../../lib/webfield-utils'
 import styles from '../../styles/components/RecruitmentForm.module.scss'
 import EditorComponentHeader from '../EditorComponents/EditorComponentHeader'
 import useTurnstileToken from '../../hooks/useTurnstileToken'
+import useUser from '../../hooks/useUser'
 
 const fieldsToHide = ['id', 'title', 'key', 'response']
 
@@ -112,6 +113,7 @@ const DeclineForm = ({ responseNote, setDecision, setReducedLoad }) => {
     allowAcceptWithReducedLoad = false,
   } = useContext(WebFieldContext)
   const [isSaving, setIsSaving] = useState(false)
+  const { user } = useUser()
   const hasReducedLoadField = invitation.edit?.note?.content?.reduced_load
   const hasCommentField = invitation.edit?.note?.content?.comment
   const fieldsToRender = orderNoteInvitationFields(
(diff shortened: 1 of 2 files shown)


melisabok:
I think "as ~Melisa_Bok1" next to Sign Up button should be removed.

xkopenreview:
> Can we catch this error and translate it to, "Profile with this email address already exists, please use the /login page"
> 
> Screenshot 2025-12-04 at 11 45 06 AM

This error has no type
I think api can update the message

carlosmondra:
> Can we catch this error and translate it to, "Profile with this email address already exists, please use the /login page"
> 
> Screenshot 2025-12-04 at 11 45 06 AM

How do they resend confirmation link to their email address? The profile is not active at this point. They cannot log in.

carlosmondra:
> Is this an API error too? @carlosmondra I think we should change it and do not show the profile that contains this email
> 
> Screenshot 2025-12-04 at 2 23 54 PM

In some cases, precreated profiles have an email address. What do we show as error message if they cannot claim profiles anymore? I am not sure what to do here. I don't think we should have more than one profile with the same email address.

We may need to remove emails from all precreated profiles... Any other ideas?

melisabok:
> How do they resend confirmation link to their email address? The profile is not active at this point. They cannot log in.

The message should say they should go to the login page and the login page has the option to resend the confirmation link. 

I don't want to reveal the profile exists and it is not active. We can say that exists and that's all or we can say that exists but it is not active yet. I prefer the first option but I'm open for the second one

racheljsmart:
Can we just delete all precreated dblp profiles?

racheljsmart:
> > How do they resend confirmation link to their email address? The profile is not active at this point. They cannot log in.
> 
> The message should say they should go to the login page and the login page has the option to resend the confirmation link.
> 
> I don't want to reveal the profile exists and it is not active. We can say that exists and that's all or we can say that exists but it is not active yet. I prefer the first option but I'm open for the second one

I think we should tell them in the error message that they should resend the link from the login page

racheljsmart:
> By showing an error if the email is taken, aren’t we still giving them some info about profile existence in the system?
> 
> If the email exists, could we email the person instead and say “if this is you, log in or reset your password. If it’s not you, no action is needed.”

I think making the process even more opaque is adding frustration to the signup process, which is already challenging. If they can't access the email and don't know the profile id, i don't know what they can really do with the knowledge the profile exists.

PR #2556, opened 2025-12-02
Fix/ Disable accessing profile by email

xkopenreview:
This is should remove call to profile or profile search by email
mainly affecting 

- /profile page
- profile search widget
- consoles
su can still access profile page and pc console as usual
su won't be assigned as reviewer/ac... so logic in other consoles are removed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,9 +35,9 @@ jobs:
 
     steps:
       - name: Start MongoDB
-        uses: supercharge/mongodb-github-action@1.10.0
+        uses: supercharge/mongodb-github-action@1.12.1
         with:
-          mongodb-version: 6.0
+          mongodb-version: 8.0
           mongodb-replica-set: rs0
       - name: Configure sysctl Limits
         run: |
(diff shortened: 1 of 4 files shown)


xkopenreview:
link to assigned sac in ac console would not work if sac is assigned using email
because can't get sac profile and can't construct link to profile using email

so only sac assigned using tilde id will be shown

components/EditorComponents/ProfileSearchWidget.js
```diff
@@ -284,8 +284,10 @@ const ProfileSearchFormAndResults = ({
     let paramKey = 'fullname'
     let paramValue = cleanSearchTerm.toLowerCase()
     if (isValidEmail(cleanSearchTerm)) {
-      paramKey = 'email'
-    } else if (cleanSearchTerm.startsWith('~')) {
+      promptError('Searching by email is not supported.')
```
melisabok: can we offer the user other ways to search in the error message?

"Searching by email is not supported, try search by profile id, name or institution domain"

melisabok:
Carlos and I decided that the API is going to replace the author id email address with a tilde id when the user signs up

melisabok:
I get this error when opening the ARR PC Console / AC tab:

```
ProgramChairConsole.js:1139 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'profiles')
    at loadSacAcInfo (ProgramChairConsole.js:1139:63)
```

xkopenreview:
>@xkopenreview is it possible to override this description if contains "email address"? at least until all the invitations are created using the new instructions

if no pc modify the default description i think it's safer to overwrite the whole default description

PR #2563, opened 2025-12-04
Fix/ PC Console - Reviewer recommendation signed with anon AC id

xkopenreview:
currently AC recommendation related calculation in PC Console assume reviewer recommendation edges added by AC are signed with tilde id of the AC

this is changed in https://github.com/openreview/openreview-py/pull/2773 where the signature is changed to paper anon ac id so the acRecommendationsCount used in overview tab and AC status tab are wrong

this pr should fix this by mapping the paper anon AC id to AC profile id and use it in acRecommendationsCount calculation


xkopenreview:
1. what's the rational of changing the signature?
2. in test of openreview/openreview-py/pull/2773, an AC reviewer recommendation edge is signed by SAC group, which would cause the progress in overview tab to be wrong.

xkopenreview:
> @xkopenreview another question - it looks like when an AC posts just 1 recommendation edge they are counted as "completed" in the stat. It looks like it should only count if they completed their required number of recommendations. Is it possible we're not checking the minReplies from the invitation?

it checks for taskCompletionCount, i think this is the min reply for tag invitations
@melisabok should it be changed to minReplies

PR #2567, opened 2025-12-05
Fix/ Add anon ac group (signature) instead of all acs group when adding missing readers

xkopenreview:
This pr should:
1. add a null check missed in #2549 which can be temp fixed by adding empty extraDisplayFields in console config and not adding option to sort when there's no display field
2. add anon AC group (the signature) instead of paper ACs group when invitation allow anon AC group

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openreview-web",
-  "version": "1.14.49",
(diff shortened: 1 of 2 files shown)


xkopenreview:
discussed about this
it's not working because it's not using inGroup: ACs

PR #2579, opened 2025-12-11
Chore: Remove nvm

carlosmondra:
We don't use nvm anymore


xkopenreview:
how about nginx
in deploy-prod
is new nginx using nvm?

PR #2580, opened 2025-12-12
Fix/ Donate - add donate page

xkopenreview:
this pr should add a donate page
and update footer to point to the donate page

### DIFF
--- a/app/Footer.js
+++ b/app/Footer.js
@@ -32,7 +32,7 @@ export default function Footer() {
                   <Link href="/sponsors">Sponsors</Link>
                 </li>
                 <li>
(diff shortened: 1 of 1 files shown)


xkopenreview:
@melisabok please take a look at latest commit

xkopenreview:
i feel it's too tightly coupled with stripe
i thought later papal venmo will be added too

xkopenreview:
> > it would also remove the thank you message after success direction correct?
> > I don't want to remove it but show it once.
> > Every time I refresh the page, it is being shown.

[7de8982](https://github.com/openreview/openreview-web/pull/2580/commits/7de8982cb0951803156f78e38e4185698210eb2c)

PR #2588, opened 2025-12-17
Registration Forms: Allow loading from different domains

haroldrubio:
This PR adds support for cross-venue registration forms and tasks in the Program Chair and Console views. It supports fetching and displaying registration data (forms, notes, and tasks) from different domains based on the Webfield config

**Changes**
`ConsoleTaskList.js`
- Support `registrationFormDomainMap` in the Webfield context which has the form:
```
registrationFormDomainMap: {
      '/Area_Chairs/-/Registration': 'aclweb.org/ACL/ARR',
      '/Area_Chairs/-/Max_Load_And_Unavailability_Request': 'aclweb.org/ACL/ARR',
      '/Area_Chairs/-/License_Agreement': 'aclweb.org/ACL/ARR',
      '/Area_Chairs/-/Emergency_Metareviewer_Agreement': 'aclweb.org/ACL/ARR'
    }
```
- Finds noteInvitations that have a matching suffix, remove the ones from the current venue and push the ones from the root venue

`ProgramChairConsole.js`
- Support `registrationFormOverrides` in the Webfield context which has the form:
```
registrationFormOverrides: {
      '/Reviewers/-/Registration': {
        domain: 'aclweb.org/ACL/ARR',
        invitation: 'aclweb.org/ACL/ARR/Reviewers/-/Registration'
      }
}
```
- Updated the console data fetching logic to retrieve registration forms and their submitted notes from the specified override domains instead of defaulting to the current venueId.
- Searches retrieved the registration form notes (ending in _Form) for invitations that have a matching suffix, swaps out the form note from the current venue and pushes the one from the root venue
- When querying registration notes, search for matching suffix again in `registrationFormOverrides` to select the right domain in the query

### DIFF
--- a/components/webfield/ConsoleTaskList.js
+++ b/components/webfield/ConsoleTaskList.js
@@ -66,10 +66,9 @@ const ConsoleTaskList = ({
       // If registrationFormInvitations is specified, filter out matching invitations
       // from domain query and use the ones from props instead
       if (registrationFormInvitations.length > 0) {
-        // Get suffixes from prop invitation IDs (e.g., 'Registration' from '.../-/Registration')
         const propSuffixes = registrationFormInvitations.map((id) => id.split('/-/').pop())
 
-        // Filter out invitations that match the prop suffixes or end with _Form, _Request, _Agreement
+        // Filter out invitations that match the prop suffixes for replacement
         allInvitations = allInvitations.filter((inv) => {
           const invSuffix = inv.id.split('/-/').pop() || ''
           return (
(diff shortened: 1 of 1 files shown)


melisabok:
Instead of passing this new config, I would like to change the current behavior that loads the registration notes:

```
      const getRegistrationFormResultsP = Promise.all(
        prefixes.map((prefix) =>
          api
            .getAll(
              '/notes',
              {
                invitation: `${prefix}/-/.*`,
                signature: venueId,
                select: 'id,invitation,invitations,content.title',
                domain: venueId,
              },
              { accessToken }
            )
            .then((notes) =>
              notes.filter((note) => note.invitations.some((p) => p.endsWith('_Form')))
            )
        )
      )
```

We should receive a list of registration form invitations as part of the component config and get the notes for every invitation in the array. That would solve the issue with ARR having another invitation id for the registration form.

PR #2596, opened 2025-12-19
Fix/ Donate - add in banner and confirm registration modal

xkopenreview:
branched off #2554 
add donate link to open open banner and confirm registration modal


xkopenreview:
the only new change in this pr is the consider donation in new name confirmation modal
which may not be necessary if donate is moved to profile page

PR #2603, opened 2025-12-23
Fix/ Reviewer Console - support review policy

xkopenreview:
this pr should add a new config edgeInvitationIds to reviewer console
and use the invitation ids defined in it to load edges and display the label/weight in header

this pr should also update loading of  custom max papers edges to use profile.id

### DIFF
--- a/components/webfield/ReviewerConsole.js
+++ b/components/webfield/ReviewerConsole.js
@@ -660,7 +660,7 @@ const ReviewerConsole = ({ appContext }) => {
     // #region get reviewer edges
     const getReviewerEdgesP = edgeInvitationIds.length
       ? Promise.all(
-          edgeInvitationIds.map((invitationId) =>
+          edgeInvitationIds.flatMap((invitationId) =>
             api
               .get(
                 '/edges',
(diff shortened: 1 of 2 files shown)


xkopenreview:
> When I run this with the ICML tests where the edges look like this:
> 
> ```
> invitation = 'ICML.cc/2023/Conference/Reviewers/-/Review_Policy',
> head = 'ICML.cc/2023/Conference/Reviewers',
> tail = '~Reviewer_ICMLOne1',
> signatures = ['ICML.cc/2023/Conference/Program_Chairs'],
> label = 'Policy A'
> ```
> 
> It displays the invitation name, but not the label: Screenshot 2026-01-05 at 4 27 34 PM
> 
> Am I missing something with the config? I have this: `edgeInvitationIds: ['ICML.cc/2023/Conference/Reviewers/-/Review_Policy']`

i think either the invitation id or the invitation used to create the edge you specified is wrong
one is icml.cc one is icml.org

enrubio (approved):
This looks great! Thank you Xukun.

Screenshot 2026-01-06 at 9 53 55 AM

If there are multiple edges, it can look like this for example:
`Review Policy: Policy A, Policy B, 0.4`

PR #2609, opened 2026-01-12
Add export all my reviews function.

leoleoasd:
Fixes openreview/openreview#375.

Basically, this gives the same information as in [here](https://docs.openreview.net/how-to-guides/data-retrieval-and-modification/how-to-get-all-the-reviews-that-i-have-written-and-their-corresponding-submissions) but in the form of UI (that users can easily save to PDF) instead of a JSON.

### DIFF
--- a/app/(Home)/NavUserLinks.js
+++ b/app/(Home)/NavUserLinks.js
@@ -54,6 +54,9 @@ export default async function NavUserLinks() {
           <li>
             <Link href="/profile">Profile</Link>
           </li>
+          <li>
+            <Link href="/reviews">All My Reviews</Link>
+          </li>
           <li className="visible-sm-block">
             <Link href="/notifications" prefetch={false}>
               Notifications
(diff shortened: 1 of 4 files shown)


app/(Home)/NavUserLinks.js
```diff
@@ -54,6 +54,9 @@ export default async function NavUserLinks() {
           <li>
             <Link href="/profile">Profile</Link>
           </li>
+          <li>
```
melisabok: I would prefer to show this option inside the activity page instead of an option in the menu. Not all the OpenReview users are reviewers.

leoleoasd: Great suggestion!

leoleoasd: Where on the activity page would be better? Maybe I can create a tab-like style, so one tab would be existing activity and another tab would be "reviews history"?

PR #2616, opened 2026-01-27
Fix/ Read user token

xkopenreview:
related to https://github.com/openreview/openreview-api/pull/937
where access token is changed to http only meaning it is not available in client code running in browser (still available in server rendered code and middleware)

so this pr should read user token instead of access token to check user info including whether user has logged in
user token has less info compared to access token so in some places where email is required, the full profile is fetched

api authentication will rely on cookie set in header instead of setting access token in authentication header

#2029 is reverted as api.get is passed on the server so it can't be authenticated
api.get is not used by any webfield

### DIFF
--- a/unitTests/NoteEditor.test.js
+++ b/unitTests/NoteEditor.test.js
@@ -157,13 +157,9 @@ describe('NoteEditor', () => {
       'token'
     )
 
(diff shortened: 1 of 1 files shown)


melisabok (approved):
Merging this to run final tests in the dev site

PR #2625, opened 2026-01-30
Fix/ EIC task list - remove link and update style

xkopenreview:
this pr should update eicTaskList used by ae console and eic console to
- remove task link to forum page
- style the text and completed tasks to match style of tasks page. existing code is adding a "completed" class which does not correspond to any style so it's removed

### DIFF
--- a/app/forum/Forum.module.scss
+++ b/app/forum/Forum.module.scss
@@ -20,17 +20,16 @@ $gray3: rgba(0, 0, 0, 0.05);
         line-height: 1.75rem;
         button {
           margin-left: 0.375rem;
+          &.expired {
+            background-color: $gray3;
+            color: gray;
+            border-color: gray;
+            text-decoration: line-through;
+          }
         }
       }
       .top-level-invitations {
         margin-top: 1rem;
-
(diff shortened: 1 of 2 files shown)


client/webfield-v2.js
```diff
@@ -1124,16 +1124,14 @@ module.exports = (function () {
...
       return (
-        '<li class="note ' +
-        (inv.complete ? 'completed' : '') +
-        '">' +
-        '<h4><a href="/forum?id=' +
-        inv.forumId +
-        (inv.complete ? '' : '&invitationId=' + inv.id) +
-        (options.referrer ? '&referrer=' + options.referrer : '') +
-        '" target="_blank">' +
+        '<li class="note"' +
```
melisabok: all the consoles of the journal are using this code to render their assigned tasks so the tasks won't be clickable for them. 

I think we need to pass some parameter in the AE console paper status tab to indicate that we do not want to render a link.

melisabok: I discussed it with Xukun and he explained this code is not being used for the tasks tab. It is good to merge now.

PR #2638, opened 2026-02-06
stop getting profiles in old forum page

melisabok:


carlosmondra:
> @carlosmondra can we consider stop returning the `tauthor` for API1 venues?

For every request or for which requests?

melisabok:
Tests are passing 🙏💪🙌

PR #2639, opened 2026-02-06
Feature/mfa

xkopenreview:
web changes for https://github.com/openreview/openreview-api/pull/938

rollout process require some thinking
extra changes required when opt in mfa is encouraged/enforced/strictly enforced
such as showing a banner or changing profile status


carlosmondra:
Wow! It works pretty well already. I just have a couple of comments:

Some typos here:
Recovery codes is the last resort of accessing your account when you lose access to all other verification methods.Please save the following recovery codes in a safe place.
You will NOT be able to login if you lost them. The recovery codes will NOT be shown again.
I suggest we change it to:
Recovery codes are your last way to access your account if you lose all other verification methods.
Save these codes in a secure place.
If you lose them, you will not be able to log in.
These recovery codes will not be shown again.

When I log out, the UI displays the error: profile not found error.

I wonder if the error message: “Account temporarily locked due to too many failed MFA attempts” should also include when the account will be unlocked. I think this is an API change, but I want to discuss it here. The current lock out time is 15 min.

racheljsmart:
I didn't notice the Password and Security tab when I clicked on Edit Profile, I had to look at Melisa's screenshot to locate where I was supposed to configure MFA. Is there a way we can indicate the tab is unread to get the user's attention for the first time?

PR #2642, opened 2026-02-09
Fix/ Reply note reader - handle parent reader with inGroup

xkopenreview:
this pr should handle the following case:

a reply to a forum reply where 
parent reader is not specified in reply invitation
the parent reader is member of a inGroup item specified in reply invitation
but the user can't see the group of the inGroup group specified

this pr should add the parent reader to options

an example is
a reviewer replying to a comment posted by AC_abcd
the comment is only visible to reviewers and AC_abcd, not visible to ACs group
reviewer replying using an invitation which specified inGroup: ACs group and AC_abcd is member of ACs group
reviewer do not have access to ACs group

this pr should add the parent reader AC_abcd to reader options

this pr should also update som variable name so it's readable

### DIFF
--- a/unitTests/NoteEditorReaders.test.js
+++ b/unitTests/NoteEditorReaders.test.js
@@ -3780,6 +3780,16 @@ describe('NewReplyEditNoteReaders', () => {
                   inGroup: 'ICML.cc/2023/Conference/Submission1/Area_Chairs',
                   optional: true,
                 },
+                {
+                  // not related to inGroup logic tested, added to so that test is logical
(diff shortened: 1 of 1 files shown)


xkopenreview:
the change in this pr requires some form of "rule" to know that
submission1/Area_Chair_{anonid} 
is a member of 
submission1/Area_Chairs

current implementation compare that
submission1/Area_Chair_{anonid} starts with submission1/Area_Chair~~s~~

i think we briefly mentioned this in the support meeting that whether 
submission1/Secondary_Area_Chairs
is always a member of
submission1/Area_Chairs 
is vague

PR #2654, opened 2026-02-18
Fix/ Message members modal - handle object signature

xkopenreview:
right now message members modal assume the message invitation.message.signature is a string

the assumption does not hold any more and it's possible to specify invitation.message.signature as enum

this pr should show in message members modal a field to capture message signature value (if not const)


celestemartinez (approved):
Works! Thanks Xukun

PR #2659, opened 2026-02-20
Fix/ Author activation status - Show more status

xkopenreview:
email author won't be validated so this pr should update the activation status to provide more info
based on 
 - author id is tilde id or email
 - whether author has a profile
 - whether the profile is activated

and the following can happen

1. is email - won't validate
2. is tilde id but can't get any profile - profile is removed
3. is tilde id and can get profile but profile is not active - profile is inactive or rejected or pending moderation
4. profile is active

### DIFF
--- a/components/webfield/NoteSummary.js
+++ b/components/webfield/NoteSummary.js
@@ -52,7 +52,7 @@ const NoteSummary = ({
       return (
         <Icon
           name="minus-sign"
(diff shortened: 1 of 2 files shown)


haroldrubio (approved):
Changes look good! tested it by posting note edits to modify the content.authorids.

PR #2667, opened 2026-03-03
Fix/ Moderation Venue Requests - split deployed venues and venues pending deployment

xkopenreview:
this pr should change venue request tab of moderation to 2 tabs

venue requests tab - the venue request that is pending deployment, sorting logic and comment showing logic is the same as existing implementation
venues tab - shows all deployed venues, it will always show the latest reply and show the posting date of latest reply

### DIFF
--- a/app/user/moderation/(VenueRequests)/NewVenueRequestCount.js
+++ b/app/user/moderation/(VenueRequests)/NewVenueRequestCount.js
@@ -15,7 +15,7 @@ export default function NewVenueRequestCount({ children }) {
             select: 'content.venue_id',
           },
           {
-            invitation: `${process.env.SUPER_USER}/Support/Venue_Request`,
+            invitation: `${process.env.SUPER_USER}/Support/Venue_Request/-/Conference_Review_Workflow`,
             select: `parentInvitations,content.venue_id`,
           },
           { includeVersion: true }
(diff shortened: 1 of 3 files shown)


xkopenreview:
discussed this with Melisa
the sorting of venues list (deployed) is 
venues with no comment first (sorted by cdate desc)
followed by venues with comment (sorted by cdate of latest comment desc)

styles/global.scss
```diff
@@ -1049,3 +1049,22 @@ header {
...
+
+  .ant-badge-count {
+    background-color: #777 !important;
+    box-shadow: none;
+  }
+}
+
+.ant-tabs-tab-active .ant-tabs-tab-btn {
+  color: constants.$subtleGray;
+}
```
xkopenreview: ThemeProvider is added to avoid global style overwrite, it provides the "OpenReview" style of antd components
if a component needs a specific customization other than the one defined in theme provider i think it should be passed to styles or classes prop of that component
global.scss will be removed with bootstrap and jquery

what style are you trying to match?

melisabok (approved):

melisabok:
maybe we can sort the venues by recent comments from PCs, for the other venues that do not have unaswered comments from PCs, we can show the regular worklfow messages, it could be helpful to see the status/activy of the venue. 

This PR is also showing the status message

PR #2669, opened 2026-03-05
Fix/ Enhance moderation access control with token validation and support group check

xkopenreview:
instead of checking su
this pr should change the access control check of moderation page to support group

### DIFF
--- a/components/Dropdown.js
+++ b/components/Dropdown.js
@@ -142,7 +142,6 @@ export default function Dropdown(props) {
   }
   const filterOption = createFilter(props.filterOption ?? defaultFilterOption)
 
-
   if (props.hideArrow) {
     props = {
       ...props,


app/user/moderation/page.js
```diff
@@ -1,12 +1,40 @@
...
 import ErrorDisplay from '../../../components/ErrorDisplay'
+import api from '../../../lib/api-client'
+
+const errorMessage = 'Forbidden. Access to this page is restricted.'
 
 export default async function page() {
-  const { user } = await serverAuth()
-  if (!isSuperUser(user))
-    return <ErrorDisplay message="Forbidden. Access to this page is restricted." />
+  const { user, token } = await serverAuth()
```
melisabok: I would like to keep having access to the moderation page as super user

xkopenreview: to keep su access, su should be added to the support group
for su, id may be different from profile id, id should be added

@carlosmondra i suppose you have SU in the support group. let me know if that's not the case

PR #2692, opened 2026-03-16
Fix/ PC Console AC status tab - allow function definition in query search

xkopenreview:
this pr should allow 
>areaChairStatusPropertiesAllowed

pc console config to allow function search
similar as in paper status tab


xkopenreview:
> can you do the same for SAC tab?

what kind of search/filter will be performed in SAC status tab?
there's really not much info to search/filter unless it's SAC paper direct assignment that's why query search is currently disabled

haroldrubio (approved):
Changes have been working good in my testing - related openreview-py PR for ARR: https://github.com/openreview/openreview-py/pull/2975

PR #2735, opened 2026-04-09
Fix/ Workflow invitations - Update workflow invitation check condition

xkopenreview:
this pr should update the invitation filtering logic for the case that all workflow and sub invitations are filtered out
which lead to infinite loop


melisabok (approved):

celestemartinez:
@xkopenreview in which case would all invitations be filtered out?

xkopenreview:
Celeste found that the venue deployment failed so no invitation is created
not related to the exclusion invitation logic

changes in this pr does not harm

celestemartinez (approved):

melisabok (approved):

emily-grabowski (approved):

PR #2744, opened 2026-04-16
Feat/ Add /debug/ip route for verifying proxy chain

carlosmondra:
Adds a support-gated /debug/ip route that mirrors the API's /debug/ip endpoint and returns both the nextjs view (incoming x-forwarded-for, x-real-ip, host, resolved client IP) and the nodejs view obtained by proxying through the api client with remoteIpAddress set. Lets a single request from a browser validate the full proxy chain: that nginx->nextjs forwards the real client IP and that nextjs->nodejs preserves it.

Reuses existing patterns (serverAuth + checkAccess with super-user bypass, api-client remoteIpAddress option). Jest test covers auth gating, combined payload, XFF forwarding, header fallbacks, and nodejs error shaping.

### DIFF
--- a/app/debug/ip/route.js
+++ b/app/debug/ip/route.js
@@ -1,0 +1,52 @@
+import { headers } from 'next/headers'
+import serverAuth from '../../auth'
+import { checkAccess } from '../../../lib/server-access-control'
+import api from '../../../lib/api-client'
+
+export async function GET() {
+  const { user, token } = await serverAuth()
+
+  const hasAccess = await checkAccess(
+    user,
+    token,
+    'DebugIp',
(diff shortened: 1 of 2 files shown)


xkopenreview:
what's the background of this change?

carlosmondra:
> what's the background of this change?

If you mean the reason for this change is that I want to make sure that the API gets the correct IP address of the client. Given that the request will go through various proxies and a new one will be added (Google's external load balancer), I want to have a way to check that the IP address of the client is not lost and it's correctly resolved when the nextjs servers make requests.

xkopenreview:
discussed this in the meeting
decided to close

PR #2746, opened 2026-04-20
Fix/ Navbar - rewrite with antd and CSS modules

xkopenreview:
## Summary
- Replace Bootstrap navbar with antd `Dropdown` (tablet/desktop) and `Drawer` (mobile);
- split into different components for different screen sizes for clarity and avoid layout shift
- moved bootstrap styles to legacy-bootstrap-nav.module.scss


enrubio (commented):
I'm testing it in the UI. 
1. I noticed that if you have a smaller window, have the dropdown open, then resize to a larger window, the dropdown doesn't resize with it:
- Medium -> Large:
Image

- Mobile -> Large:
Image

It works correctly on the live site right now, so I think we should fix it.

2. Possibly minor, but the current mobile navbar has a clickable dropdown on your profile ID that leads to Profile, Password & Security, Logout. The new mobile navbar has no dropdown and displays all the options.
- I think we either have it function the same or we indent the Profile & Password tabs. Otherwise, the profile ID looks clickable but it's not:

Image

xkopenreview:
> I'm testing it in the UI.
> 
> 1. I noticed that if you have a smaller window, have the dropdown open, then resize to a larger window, the dropdown doesn't resize with it:
> 
> * Medium -> Large:
> 
> Image
> * Mobile -> Large:
> 
> Image
> It works correctly on the live site right now, so I think we should fix it.
> 
> 2. Possibly minor, but the current mobile navbar has a clickable dropdown on your profile ID that leads to Profile, Password & Security, Logout. The new mobile navbar has no dropdown and displays all the options.
> 
> * I think we either have it function the same or we indent the Profile & Password tabs. Otherwise, the profile ID looks clickable but it's not:
> 
> Image

thanks
please take a look at [0e85e6a](https://github.com/openreview/openreview-web/pull/2746/commits/0e85e6adda4376555c7892ecc102167c5cec317f)

enrubio (approved):
The changes look great, it's very smooth. I don't notice any other issues in the UI.

melisabok:
can the border be removed when the element is selected?

Screenshot 2026-04-27 at 10 49 45 AM

xkopenreview:
> can the border be removed when the element is selected?

i think this is added by chrome
should be fixed by [86caf86](https://github.com/openreview/openreview-web/pull/2746/commits/86caf86e17ff365ade4b240d9096f6cfee15f397)

melisabok (approved):

PR #2753, opened 2026-04-24
Feature/multiple roles

melisabok:
Show multiple reviewer roles in the timeline UI


xkopenreview (approved):

celestemartinez (approved):
I checked and it seems like we are showing all the invitations and groups we should be showing.

One comments though:
1. The PC group is now shown at the bottom of the workflow groups. I don't necessarily think this is an issue but I was used to clicking on it to get to the PC console. Not sure if that was on purpose.

PR #2756, opened 2026-04-27
Remove isomorphic-dompurify

### LINKED ISSUE #2742: Remove isomorphic-dompurify
looks like isomorphic-dompurify can be replaced with just dompurify

_Originally posted by @xkopenreview in https://github.com/openreview/openreview-web/issues/2741#issuecomment-4260648238_

xkopenreview:
this pr should replace isomorphic-dompurify with dompurify
and fix the version to 3.4.1 (current latest)


components/forum/ForumNote.js
```diff
@@ -92,7 +90,7 @@ function ForumNote({ note, updateNote, deleteOrRestoreNote }) {
 
       <ForumTitle
         id={id}
-        title={DOMPurify.sanitize(content?.title?.value)}
+        title={content?.title?.value}
```
xkopenreview, line 93: this is added after the security audit
but looks like it's not necessary because it will be shown as string

melisabok: I don't understand this comment.

melisabok: why do you need to remove it?

xkopenreview: the old package works on both server rendered page and client rendered page
new package works only on client rendered page

the ForumNote is server rendered so the new package would not work
but the title is rendered as string no matter what the title is
so the sanitize actually has no effect

carlosmondra (approved):

melisabok (approved):

PR #2763, opened 2026-05-01
Fix/ Forum page access with missing User-Agent header

xkopenreview:
when forum page is accessed with no user-agent, the page may throw server error checking user-agent
it should not happen for human user and most if not all real crawlers

this pr adds a check so that it does not throw server error

## Summary
- Guard `userAgent.includes('Googlebot')` against a null User-Agent header in `app/forum/page.js`. Requests with no UA (bare scripts, non-browser clients) were throwing `TypeError` on the catch path, causing the page to error instead of falling through to the intended login redirect / permission-error response.
- Adds `userAgent &&` so a missing UA falls through to the permission-error response. Real crawlers (Googlebot, Bingbot, BingPreview, Safe Links) all send identifying UAs, so this doesn't affect indexing.


app/forum/page.js
```diff
@@ -1,20 +1,21 @@
...
 import CommonLayout from '../CommonLayout'
-import { referrerLink, venueHomepageLink } from '../../lib/banner-links'
-import Banner from '../../components/Banner'
-import legacyStyles from './LegacyForum.module.scss'
-import LegacyForum from '../../components/forum/LegacyForum'
-import ErrorDisplay from '../../components/ErrorDisplay'
 import ArxivForum from './ArxivForum'
 
+import styles from './Forum.module.scss'
+import legacyStyles from './LegacyForum.module.scss'
```
carlosmondra, line 17: what are all these changes for?

xkopenreview: sortImports by oxfmt
didn't do it for all files

so when a file is changed, the imports will be sorted.
sort rule is in sortImports in .oxfmtrc.json, roughly

normal imports
-- empty line --
styles import of css module 
import of legacy bootstrap style

carlosmondra (approved):

PR #2764, opened 2026-05-01
Fix/ Open reply forum break mathjax rendering

xkopenreview:
when there's a forum reply with formula, adding a reply to this formula reply will cause it to lose formula rendering

this is caused by the structure of ForumReply component cause the formula reply to be re-rendered when active reply invitation is set to show note editor thus the formula rendering is lost

this pr should fix the issue by creating a new component ForumReplyForm which contains add reply buttons + reply note editor so that setting invitation does not cause the parent reply to re-render

this pr should also change ForumReplyForm to use antd and fix a few minor issues:
1. when there are multiple add reply buttons, for example review and comment, clicking on review button open the review form, if user now click comment button, it will close the form
with this pr, it will switch to comment form instead
2. when add button is at bottom of page, the note editor will be shown off screen
this pr should scroll to the add button so that the note editor at least top part of the note editor is visible


melisabok (approved):

enrubio (approved):
Changes look good

PR #2767, opened 2026-05-06
Immutable web deploy

tyler-dominguez:
## Summary

Replaces the SSH-based mutable deploy with an immutable golden image pipeline.

**Deploy flow:**
1. Build Next.js artifact and upload to GCS (skipped if build already exists for that tag + config hash)
2. Packer bakes the artifact into a new GCE golden image on top of `openreview-base`
3. Terraform updates the regional instance template and MIG configuration
4. `gcloud rolling-action start-update` rolls the MIG zero-downtime (max-surge=3, max-unavailable=0)
5. Nginx static file deploy runs after rollout completes
6. Image name is recorded to GCS (`web-image-map/<tag>`) for rollback

**Rollback flow (`rollback: true` input):**
- Dispatch with a tag and check `rollback` — skips build, Packer, and Terraform entirely
- Resolves the tag to its baked image via GCS mapping and rolls the MIG directly (~1 min)

## Notes
- Rollback mapping is only populated for deploys made after this merge; backfill manually with `echo "<image>" | gsutil cp - gs://openreview-general/web-image-map/<tag>`
- Connection drain timeout on `openreview-web-backend-service` is 300s — tracked separately for reduction

## Test plan
- [x] End-to-end deploy validated with `v1.15.16` — all steps passed in 13m30s — [job](https://github.com/openreview/openreview-web/actions/runs/25400184095/job/74497356438)
- [x] Rollback validated with `v1.15.16` — skipped build/Packer/Terraform, completed in 1m02s — [job](https://github.com/openreview/openreview-web/actions/runs/25441554234/job/74633934502)


.github/workflows/deploy-prod.yml
```diff
@@ -1,20 +1,34 @@
...
       tag:
         description: The branch, tag, or commit hash to deploy
         required: false
         default: master
         type: string
+      rollback:
+        description: "Roll back to a previously deployed tag (skips build and Packer)"
+        required: false
+        default: false
+        type: boolean
```
carlosmondra: Could we just pass the tag of the previous version instead?

.github/workflows/deploy-prod.yml
```diff
@@ -29,6 +43,25 @@ jobs:
         with:
           ref: ${{ env.TAG }}
           fetch-depth: 0
+
+      # Requires a GitHub App installed on openreview/openreview-cloud with contents:read permission.
+      # Set APP_CLIENT_ID and APP_PRIVATE_KEY secrets on this repo before running.
```
carlosmondra: We can remove these comments :)

.github/workflows/deploy-prod.yml
```diff
@@ -75,35 +113,82 @@ jobs:
           echo "BUILD_ID=\"$build_id\"" >> ./deployment.conf
           gsutil cp ./deployment.conf gs://openreview-general/conf/deployment.conf
 
-          instance_prefix='openreview-web-'
-
-          instances=$(gcloud compute instances list | grep "$instance_prefix" | grep RUNNING | tr -s ' ' | cut -d' ' -f1,2)
+      - name: Setup Packer
+        if: ${{ !inputs.rollback }}
+        uses: hashicorp/setup-packer@main
```
carlosmondra: There is no versioning here?

tyler-dominguez: Will pin.

.github/workflows/deploy-prod.yml
tyler-dominguez: much cleaner will add.
Discussed niche where some niche strategies would have a problem with this. Does not effect us.

.github/workflows/deploy-prod.yml
```diff
@@ -75,35 +117,82 @@ jobs:
...
-          nvm='export NVM_DIR="$HOME/.nvm" && [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" &&'
-
           nginx=$(gcloud compute instances list | grep "$nginx" | grep RUNNING | tr -s ' ' | cut -d' ' -f1,2)
 
           nginx_arr=(${nginx// / })
           nginx_name=${nginx_arr[0]}
           nginx_zone=${nginx_arr[1]}
 
-          gcloud compute --project "sunlit-realm-131518" ssh --zone $nginx_zone --tunnel-through-iap openreview@$nginx_name --command "${nvm} bash bin/deploy-web.sh ${TAG}" --quiet
+          gcloud compute --project "${{ secrets.GCP_PROJECT_ID }}" ssh --zone $nginx_zone --tunnel-through-iap openreview@$nginx_name --command "${nvm} bash bin/deploy-web.sh ${TAG}" --quiet
```
carlosmondra, line 210: I think we need to check what deploy-web.sh does in nginx. I don't remember if it does a build if it cannot find the build in the bucket. This is necessary so that the static files are served by nginx. We can remove it once we start serving them from the CDN

tyler-dominguez: [#61 - Addressing Here](https://github.com/openreview/openreview-cloud/pull/61)

tyler-dominguez: [#64 - Further addressed here](https://github.com/openreview/openreview-cloud/pull/64)

tyler-dominguez: Modernized deploy-web.sh on ngnix for statics still in use in the meantime.

carlosmondra:
When the github actions script finishes running, it will still be doing the rolling update, correct?

tyler-dominguez:
> When the github actions script finishes running, it will still be doing the rolling update, correct?

No, Github action waits for all MIGs to be restarted before completing.

carlosmondra (approved):

tyler-dominguez:
https://github.com/openreview/openreview-web/actions/runs/25741367118/job/75592355660

Current run

carlosmondra:
> https://github.com/openreview/openreview-web/actions/runs/25741367118/job/75592355660
> 
> Current run

We are getting:
```
AccessDeniedException: 403 nginx-reverse-proxy@***.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist).
```
 I suggest that github actions should be the one copying the html files to the bucket even if nginx continues to serve some files. You may need to grant the bucket write permission for the nginx-reverse-proxy SA

xkopenreview:
i have two questions:
1. what will be the process of deployment when there's no code change but the change in env var in cloud storage?
for example the feature was deployed but disabled by a flag in env var, the only change to enable this feature is to change the config
2. sometimes master branch is deployed without tag, it looks like subsequent master deployments will all refer to the first master deployment?

carlosmondra (approved):

tyler-dominguez:
https://github.com/openreview/openreview-web/actions/runs/25751830761/job/75629892700

Completed with force rebuild

xkopenreview:
i think maybe force rebuild should default to true
for regular release it will have no effect setting it to true
for re-deployment i think it's worse releasing outdated code compared to creating duplicated image 

---
discussed about this
rebuilding the image is taking too long so it's better to remember to check the checkbox when force build is needed

xkopenreview (approved):

PR #2770, opened 2026-05-07
Fix/ Remove rc-picker

xkopenreview:
built on top of #2768 

this pr should replace rc-picker with antd datetimepicker (used rc-picker underneath)
tests added in #2768 should still pass


components/DatetimePicker.js
```diff
@@ -47,23 +46,46 @@ const DatetimePicker = ({
...
-      onChange={handleChange}
-      onBlur={onBlur}
-      placeholder={placeholder ?? 'Select datetime'}
-      use12Hours
-      autoFocus={autoFocus}
-      allowClear={allowClear}
-      disabledDate={disabledDate}
-      getPopupContainer={getPopupContainer}
-    />
+    <ConfigProvider
```
xkopenreview, line 52: A local config provider is added to avoid polluting global theme provider and global classname overwrite (.ant-)
and to mimic existing style

melisabok (approved):

xkopenreview:
replaced by #2836

PR #2771, opened 2026-05-08
Fix/ Password reset - allow user to choose log out all sessions

xkopenreview:
related to https://github.com/openreview/openreview-api/pull/1099

this pr should add a checkbox in password reset page /user/password to allow user to logout from other session if user is logged in

this pr should also remove bootstrap from /user/password page
this pr should also rename the form where user enter the email to request password reset for from
ResetForm
to
PasswordResetRequestForm 
to avoid confusion


carlosmondra (approved):
Tested it on dev site and it works. The UI in the other devices does not refresh, but I think that's OK.

xkopenreview:
> The UI in the other devices does not refresh, but I think that's OK.

it should show something when api call is made or page reloaded

carlosmondra:
> > The UI in the other devices does not refresh, but I think that's OK.
> 
> it should show something when api call is made or page reloaded

The browser where I did the password reset reloaded correctly.
The device where I was logged out which was my phone using the Brave browser was not refreshed. But I think that's fine.

xkopenreview:
@melisabok i think api is in

PR #2803, opened 2026-05-26
Use GITHUB_WORKSPACE for deployment.conf path

tyler-dominguez:
Fixes deployment.conf path in the Run deploy script step. When a new build is created, `cd /tmp` shifts the working directory before deployment.conf is written, causing `./deployment.conf` to resolve to `/tmp/deployment.conf` instead of `$GITHUB_WORKSPACE/deployment.conf`. This means the Record image→tag mapping step uploads the wrong file to GCS.

Uses explicit `$GITHUB_WORKSPACE/deployment.conf` throughout to avoid the path dependency on working directory.


tyler-dominguez:
Closing — most changes already in master. Will re-open with just the GITHUB_WORKSPACE deployment.conf fix.

PR #2812, opened 2026-06-09
Fix/ Profile - handle Independent Researcher position in dropdowns and disable fields accordingly

xkopenreview:
## Summary

Handles the **Independent Researcher** position in the profile editor. Extracted as a standalone PR from `fix/profile-edit-vouch-changes` (single atomic commit, no vouch/relations logic).

When a user's education-history position is set to `Independent Researcher` or any string containing independent:
- The institution domain auto-fills to `independent-researcher.org` / name `Independent`.
- The relevant institution fields are disabled (`EducationHistorySection.js`).
- Department/institution handling in `Profile.js` accounts for the independent-researcher case.
- Adds a `clientOnly` prop to `CreatableDropdown` (`Dropdown.js`) so the inline dropdown renders/focuses immediately instead of flashing a loading spinner; used by the position/relation dropdowns.

independent-researcher.org can later be used for independent researcher related update, for example a page listing research related submissions/tags

### DIFF
--- a/app/profile/Profile.js
+++ b/app/profile/Profile.js
@@ -26,9 +26,14 @@ export default async function Profile({
     return (
       <div>
         {currentHistories?.map((history, index) => {
+          const isIndependentResearcher = history.position === 'Independent Researcher'
           const posititon = upperFirst(history.position).trim()
-          const department = history.institution.department?.trim()
-          const institutionName = history.institution.name?.trim()
+          const department = isIndependentResearcher
+            ? undefined
+            : history.institution.department?.trim()
+          const institutionName = isIndependentResearcher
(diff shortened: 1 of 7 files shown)


xkopenreview:
may need to add the Independent Researcher option in py test and position option
or allow user to type

if we want user to type may need to match more keyword apart from Independent Researcher

xkopenreview:
@tyler-dominguez 
~~independent.openreview.net need to point to probably openreview.net
when this is deployed~~

after discussion, decided to use independent-researcher.org instead

carlosmondra (approved):
Looks good!

melisabok (approved):

PR #2817, opened 2026-06-15
Fix/ Moderation - add domain search

xkopenreview:
this pr should allow domain (email domain + history domain) search in moderation page
this pr should also fix the display of results from multiple search keywords
- currently it just show all search results because searching by name/tilde id would not return many results
- but for domain search it can return many results so the pagination will do a local slice before showing the result

one potential problem is that when searching for names in the form of a valid domain, the search query will be wrong
if it become a real concern then need to implement a search category dropdown to checkbox to clearly indicate the search category intention instead of guessing based on search term

related to https://github.com/openreview/openreview-api/pull/1147


carlosmondra (approved):

xkopenreview:
@deanaOR do you want to take a look?
api change is already in

PR #2818, opened 2026-06-16
Forward guest challenge-clearance token through SSR and client API calls

carlosmondra:
## Summary
Mitigates guest scraping by integrating the API's Turnstile challenge flow into the web app: SSR requests forward the challenge-clearance cookie so the API gate can verify cleared guests, and guests hitting a challenge are redirected to the API-served challenge page that returns them to where they were.

## Changes
- **`lib/api-client.js`**: Forward the clearance cookie on SSR calls via a `clearanceToken` option (threaded through `getById` / `getNoteById`); on client-side `ChallengeRequiredError` (403), navigate the browser to the API challenge page with a return path built from the current URL.
- **`app/auth.js`**: `serverAuth()` reads the clearance cookie and returns `clearanceToken` for both authenticated and guest sessions.
- **`app/forum/page.js`**: Thread `clearanceToken` into forum note fetches; on `ChallengeRequiredError`, signal `challengeRequired` and `redirect()` guests to the API challenge page (with metadata falling back gracefully).
- **`.env.example`**: Add `CLEARANCE_COOKIE_NAME` (shared with the API).

### DIFF
--- a/app/auth.js
+++ b/app/auth.js
@@ -34,8 +34,10 @@ export default async function serverAuth() {
 
   const payload = getTokenPayload(token?.value)
   if (!payload || !payload.user?.profile?.id) {
+    // Guests forward the clearance token so the API gate can verify them.
     return { clearanceToken }
   }
 
-  return { token: token.value, user: payload.user, clearanceToken }
+  // Logged-in users bypass the challenge gate, so they never need clearance.
+  return { token: token.value, user: payload.user }
 }
--- a/app/auth.js
+++ b/app/auth.js
@@ -28,11 +28,14 @@ export function getTokenPayload(token) {
 export default async function serverAuth() {
   const cookie = await cookies()
   const token = cookie.get(process.env.ACCESS_TOKEN_NAME)
+  // Guest scraping mitigation: forward the challenge-clearance cookie on SSR
+  // fetches even for guests, so the API gate can verify cleared requests.
+  const clearanceToken = cookie.get(process.env.CLEARANCE_COOKIE_NAME || 'openreview.clearanceToken')?.value
 
   const payload = getTokenPayload(token?.value)
   if (!payload || !payload.user?.profile?.id) {
-    return {}
+    return { clearanceToken }
   }
 
-  return { token: token.value, user: payload.user }
+  return { token: token.value, user: payload.user, clearanceToken }
 }
--- a/app/auth.js
+++ b/app/auth.js
@@ -28,8 +28,6 @@ export function getTokenPayload(token) {
 export default async function serverAuth() {
   const cookie = await cookies()
   const token = cookie.get(process.env.ACCESS_TOKEN_NAME)
-  // Guest scraping mitigation: forward the challenge-clearance cookie on SSR
-  // fetches even for guests, so the API gate can verify cleared requests.
   const clearanceToken = cookie.get(process.env.CLEARANCE_COOKIE_NAME || 'openreview.clearanceToken')?.value
 
   const payload = getTokenPayload(token?.value)
@@ -38,6 +36,6 @@ export default async function serverAuth() {
     return { clearanceToken }
   }
 
-  // Logged-in users bypass the challenge gate, so they never need clearance.
+  // Logged-in users bypass the gate, so they never need clearance.
   return { token: token.value, user: payload.user }
 }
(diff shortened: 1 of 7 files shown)


app/auth.js
```diff
@@ -28,11 +28,14 @@ export function getTokenPayload(token) {
...
+  const clearanceToken = cookie.get(process.env.CLEARANCE_COOKIE_NAME || 'openreview.clearanceToken')?.value
 
   const payload = getTokenPayload(token?.value)
   if (!payload || !payload.user?.profile?.id) {
-    return {}
+    return { clearanceToken }
   }
 
-  return { token: token.value, user: payload.user }
+  return { token: token.value, user: payload.user, clearanceToken }
```
xkopenreview: i think this is not necessary
as clearanceToken is a guest only thing so it's already returned in ln37

app/forum/page.js
```diff
@@ -89,6 +91,11 @@ const getForumNote = async (
     }
     return { forumNote: note, version: 1 }
   } catch (error) {
+    // Guest scraping mitigation: signal that this guest must solve the challenge.
+    // The caller builds the redirect from the real page URL.
+    if (error.name === 'ChallengeRequiredError') {
```
xkopenreview, line 95: i think profile page is not showing because it silently failed with the challenge required error
so it will need something similar to this

accessing the profile directly using dev api is working because it can redirect to challenge correctly

app/challenge/route.js
```diff
@@ -0,0 +1,153 @@
...
+
+  // Substitute via replacement functions so "$" in values isn't interpreted by
+  // String.replace, and JSON-encode the JS values to prevent script breakout.
+  const html = TEMPLATE.replace(/\{\{SITE_KEY\}\}/g, () => siteKey)
+    .replace(/\{\{API_BASE\}\}/g, () => JSON.stringify(apiBase))
+    .replace(/\{\{REDIRECT\}\}/g, () => JSON.stringify(redirect).replace(/</g, '\\u003c'))
+    .replace(/\{\{LOGIN_URL\}\}/g, () => loginUrl)
+
+  return new Response(html, {
+    headers: { 'Content-Type': 'text/html; charset=utf-8', 'Content-Security-Policy': CSP },
```
xkopenreview: this page should have a X-Robots-Tag
so it's not indexed

app/forum/page.js
xkopenreview, line 84: shouldRedirect get the blind note and probably require the clearnanceToken too

app/auth.js
carlosmondra: Good call — removed. `serverAuth` now returns `clearanceToken` only on the guest path (the early return); the logged-in return drops it, since logged-in users bypass the gate and never need it.

app/forum/page.js
carlosmondra: Fixed on the profile page. `app/profile/page.js` was catching the gate's `ChallengeRequiredError` and rendering "Profile not found"; it now redirects to `/challenge` (mirroring this forum-page handling) and forwards the clearance token on the `/profiles` fetch so cleared guests aren't re-challenged.

app/challenge/route.js
carlosmondra: Added — the response now sends `X-Robots-Tag: noindex, nofollow`, so the verification interstitial (and its unbounded `?redirect=` variants) stays out of search-engine indexes.

app/forum/page.js
carlosmondra: Looked into this one and it turns out it isn't needed. `shouldRedirect`'s `/notes` call passes `version: 1`, which the api-client routes to `API_URL` (the legacy v1 API), not `API_V2_URL` (api-client.js:103-104). The challenge gate only runs on the v2 server, so this call never reaches the gate and v1 ignores a clearance cookie — forwarding it here would just imply this path is gated when it isn't. Leaving it as-is.

For contrast, the main `getNoteById` fetch above does forward the clearance token, since it hits v2.

xkopenreview:
currently profile edit name section loads the name removal notes even for activation
so currently activation will probably always hit the challenge

maybe it should avoid loading name deletion notes for new registration

xkopenreview:
hash(tab navigation) of the url will be dropped in the redirection
but i guess it's fine for guest

xkopenreview (approved):

PR #2819, opened 2026-06-16
Fix/ JSON editor - Update type of max length

### LINKED ISSUE #2810: Submission add field issue- MaxLength not converted to integer
When I tried to add a MaxLength to a submission field for a PC, I got this error (s/s below)

To resolve it, I had to create the field without a max length , then add it manually to the JSON

Image

xkopenreview:
this pr should change the type of 
- minLength
- maxLength
- minimum
- maximum
- maxSize

in json editor widget to integer so that type conversion can convert the string value to the type api is expecting.

this pr should also fix a typo bug in range which cause range not be to shown at all
so looks like range is not used and the type conversion logic is a bit complex so comment it out for now

one thing to notice is that api expect number for maxSize but integer for minimum/maximum 
but the type conversion will not know whether to convert to integer or float so set them all to integer

### DIFF
--- a/components/EditorComponents/ContentFieldEditor.js
+++ b/components/EditorComponents/ContentFieldEditor.js
@@ -1,15 +1,16 @@
-import { useContext, useState } from 'react'
 import { get, set, unset } from 'lodash'
+import { useContext, useState } from 'react'
+import { convertToType } from '../../lib/webfield-utils'
+import Dropdown from '../Dropdown'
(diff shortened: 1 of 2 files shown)


xkopenreview:
@emily-grabowski this is the issue raised in previous weekly meeting

carlosmondra (approved):

melisabok (approved):

PR #2827, opened 2026-06-22
Add temporary refresh-token diagnostic logging

carlosmondra:
> [!WARNING]
> **Temporary diagnostic code — do not merge.** These logs exist only to diagnose a staging issue and will be reverted once the root cause is found.

## Summary
Instruments the server-side token refresh path in middleware to diagnose a staging-only failure where the API returns `MissingTokenError: No refresh token found` even though the refresh cookie is present in the browser.

## Changes
- **`middleware.js`**: Adds three temporary log points around the `/refreshToken` call:
  - `refresh debug OUT` — cookie name, value length, and all cookies present (to catch empty/duplicate refresh cookies) before the API call.
  - `refresh debug IN` — API response status and body.
  - `refresh debug CATCH` — logs the previously-silent catch block so fetch/network errors surface.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -15,6 +15,17 @@ export async function middleware(request) {
 
   const decodedRefreshToken = getTokenPayload(refreshToken.value)
 
+  // TEMP DEBUG — remove after diagnosing staging refresh issue
+  // oxlint-disable-next-line no-console
+  console.log('refresh debug OUT', {
+    envName: process.env.REFRESH_TOKEN_NAME,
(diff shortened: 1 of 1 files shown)


carlosmondra:
Closing — root cause was an infra-side nginx misconfig on dev (the api2-dev refresh-token extraction regex matched the prod cookie name `openreview.refreshToken` instead of `openreview.dev.refreshToken`). Fixed in nginx; openreview-web needs no change. These were temporary diagnostic logs only.

PR #2829, opened 2026-06-25
Fix/ DatetimePicker - correct UTC timezone handling

xkopenreview:
one of the unit test "call onChange when user manually enter date only" fail in EU timezone
because the datetime picker is not keeping local time when it's configured to omit time part
causing the date to shift (to previous day in test) for users in UTC+ timezone


xkopenreview:
included in #2836

PR #2834, opened 2026-06-29
Fix/ Profile activate - add file upload section

xkopenreview:
related to https://github.com/openreview/openreview-api/pull/1160

this pr should also get rid of rc-steps which is included in antd


xkopenreview:
the data flow of files may require discussion:
1. in file upload section the user is still guest, so file upload (PUT to /attachment) fails
2. how the file url is added to profile obj, in process of Identity_Verification invitation, by api or by web
consideration is to enforce the maxitems limit, avoid guest abusing /attachment and recovery from error in profile info
3. whether to allow user to view the files uploaded in previous registration attempts (as guest)

---
discussed with @carlosmondra about this
- putting files
- retrieving file urls by invitation (show files uploaded in previous registration attempts)
- retrieving actual files (to preview)
will be handled by guest token similar to reviewer recruitment

files uploaded can't be removed so that maxItems can be enforced. file upload section allows deletion for files which are selected but not sent to /attachment yet though

file urls will be posted together with the rest of profile object by web to the activation call
---
discussed about this again
- it's expected that user can upload files which are not saved to the profile object, human verification serve as rate limit
- @carlosmondra to update api to allow identityDocuments to be saved with /activate call when user click on register button
- @xkopenreview to persist file upload tab so that switching tabs does not clear the files that user uploaded to the server but not saved to object yet

xkopenreview:
decided to pause the work on this

PR #2836, opened 2026-06-30
Fix/ Profile edit/activate - mandatory dob

xkopenreview:
this pr should
- remove rc-picker with a slighter cleaner version of #2770 without using a local config provider
- make profile dob mandatory
- show minor badge in profile preview modal
- include changes of #2829 
- move the mock of ResizeObserver and matchMedia to common jest setup so that it's not repeated in individual tests

requires https://github.com/openreview/openreview-api/pull/1219


xkopenreview:
@melisabok i think this is required for dob change

melisabok:
can the color of the time selection be the same as the ok button?

Screenshot 2026-08-06 at 9 43 32 PM

xkopenreview:
> can the color of the time selection be the same as the ok button?
> 

discussed this Melisa
changing the time selection color requires a local config provider just for date picker as in #2770 which is probably too much overhead so it's ok to leave it as it is now in this pr

profile dob changes will be based on changes in this pr

xkopenreview:
discussed with @carlosmondra about this
decided to add dob input (not selector to avoid user select random date) in signup page below name input so that users under 13 don't create inactive profile

the dob input in profile edit page will be disabled(or not shown) when the profile has dob

api will add validation that profile dob can't be changed to removed

xkopenreview:
@carlosmondra @melisabok 
i think this pr is ready for testing

melisabok:
Some feedback based on this screenshot:

- date of birth should use the width that the fullname textfield
- Remove the text "OpenReview profiles require an age of 13 or over. Your date of birth can NOT be changed after registration.". and add an "i" icon next to "Enter your date of birth" with a hover text "OpenReview collects your date of birth solely for age verification: accounts require a minimum age of 13, and additional protections apply to users between 13 and 18. Your date of birth is never shown publicly and cannot be modified after registration; corrections require review by the OpenReview.net Staff."

Screenshot 2026-08-19 at 9 50 16 AM

xkopenreview:
[c3f8091](https://github.com/openreview/openreview-web/pull/2836/commits/c3f8091a3bea072935072dbf0428dfeee6548fee) should:
1. update dob input size and instruction text in sign up page
2. allow user to select dob below 13 in profile edit page
3. show dob section in profile preview modal

xkopenreview:
@melisabok i think both web and api are ready

melisabok:
ok, I will start testing these changes

carlosmondra (approved):

PR #2838, opened 2026-07-05
Redirect legacy webfield guests through the API v1 scraping challenge gate

carlosmondra:
## Summary
Companion web change for the API v1 challenge gate (openreview/openreview-api-v1#3129). Legacy venue pages (V1 webfields) fetch through jQuery `$.ajax`, which bypasses the `lib/api-client.js` `checkStatus` handler — and several calls use `handleErrors: false`, so they never reach `jqErrorCallback` either. A guest blocked by the gate got a swallowed `403 ChallengeRequiredError` with no redirect and no Turnstile widget. This adds a single central hook so those guests are sent to the challenge page and returned to where they were once cleared.

## Changes
- **`client/getChallengeRedirect.js`** (new): pure helper — given a failed jqXhr and the current path, returns the `/challenge?redirect=…` URL for a `403 ChallengeRequiredError` (with a `responseText` JSON fallback), otherwise `null`.
- **`client/globals.js`**: registers one global `$(document).ajaxError` handler that fires for every legacy `$.ajax` request regardless of its `handleErrors` setting, and navigates blocked guests to the challenge page. Mirrors the client-side redirect already in `api-client.js`.
- **`unitTests/ChallengeGateRedirect.test.js`** (new): covers the redirect decision — 403 + `ChallengeRequiredError` → redirect (incl. `responseText` fallback), `ForbiddenError` / non-403 / malformed body / missing xhr → no redirect.

## Notes
Requires the API v1 gate (openreview/openreview-api-v1#3129). For the clearance cookie (set by the V2 `/challenge/verify`) to satisfy the v1 gate, it must be scoped to cover both API hosts (e.g. `Domain=.openreview.net`) and the v1/v2 gates must share the signing secret — a deployment/config matter, not web code.

### DIFF
--- a/client/getChallengeRedirect.js
+++ b/client/getChallengeRedirect.js
@@ -1,0 +1,20 @@
+// Given a failed jQuery ajax XHR and the current page path, returns the URL of
+// the Turnstile challenge page a guest should be sent to when blocked by the
+// scraping challenge gate, or null when the error is not a challenge requirement.
+// Legacy $.ajax requests bypass the api-client checkStatus handler, so
+// client/globals.js uses this to redirect blocked guests.
+module.exports = function getChallengeRedirect(jqXhr, currentPath) {
+  if (jqXhr?.status !== 403) return null
+
+  let errorName = jqXhr.responseJSON?.name
+  if (!errorName && jqXhr.responseText) {
+    try {
+      errorName = JSON.parse(jqXhr.responseText)?.name
+    } catch {
+      errorName = null
+    }
+  }
+
+  if (errorName !== 'ChallengeRequiredError') return null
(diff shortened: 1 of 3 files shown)


xkopenreview:
i thought only notes and profiles are gated?

carlosmondra:
> i thought only notes and profiles are gated?

I saw a crawler GETting invitations and I'm worried that invitations can leak information. Also, I want to protect API 1 before we do the migration.

xkopenreview:
discussed with Carlos about this
the changes in homepage will be reverted
and an exception will be added for news article invitation as currently the call to get news will silently fail for guest

xkopenreview (approved):

PR #2807, opened 2026-05-27
Fix/ Profile - Vouch

xkopenreview:
this pr contains the following changes:
1. handle display of independent researcher and independent researcher related institution change
2. fix institution/region dropdown disappearing while loading
3. handle vouch in relation section

PR #2839, opened 2026-07-06
Fix/ Remove vouch tag invitation option

xkopenreview:
with introduction of vouch #2807, the function to add vouch tag in profile preview modal is obsolete
as the vouch invitation has been updated to include relation name, start and end year instead of just the voucher id

the voucher should posting this tag instead of moderator

so this pr should keep only the general moderation label invitation

this pr should also fix the issue that the saved tag options stay open after the the tag is posted
this pr should also add a opaque state in profile tags to indicate the tags are being updated


components/profile/ProfilePreviewModal.js
```diff
@@ -26,17 +26,9 @@ const ProfilePreviewModal = ({
   const [isRejecting, setIsRejecting] = useState(false)
   const [rejectionReasons, setRejectReasons] = useState([])
   const [error, setError] = useState(null)
-  const tagInvitationOptions = [
-    {
-      label: 'General Moderation Tag',
-      value: `${process.env.SUPER_USER}/Support/-/Profile_Moderation_Label`,
-    },
-    {
-      label: 'Vouched by Tag',
```
melisabok, line 35: what if we want to vouch on behalf of the user? I would like to keep it just in case.

xkopenreview: it needs the relations name, start year and end year and we said moderator will do it with a py script

melisabok (approved):

PR #2849, opened 2026-07-17
Fix/ Profile edit page - Update vouch button text

xkopenreview:
some users report that the vouch button is not obvious
this pr should add "Vouch" text to make it more obvious


melisabok (approved):

xkopenreview:
explored alternative options to make it even more obvious but does not work very well with style of the input
so add button text first to see how it works

emilygrabowski:
I tested it and the vouch button looks much more clear. 

One comment- rather than the vouch button not appearing if the user is eligible for a vouch, could it also appear and give an error appropriate to the profile status? Such as: Profile is already active, etc.

xkopenreview:
it will make disabled vouch button to appear to many user's relations who has no intention to vouch for anyone

PR #2851, opened 2026-07-17
Speed up user moderation queue query

### LINKED ISSUE #2850: Slow query user moderation
Change `GET /profiles?limit=200&needsModeration=true&offset=0&sort=tmdate%3Adesc`
To `GET /profiles?limit=200&offset=0&sort=tmdate%3Adesc&state=Needs%20Moderation`

carlosmondra:
## Summary
The user moderation queue used a slow `needsModeration=true` filter to fetch pending profiles. This switches to the faster `state=Needs Moderation` query, addressing #2850.

## Changes
- **`app/user/moderation/UserModerationTab.js`**: Replaced `needsModeration: true` with `state: 'Needs Moderation'` in the moderation profiles query. Safe because the profile-state dropdown that also sets `state` only renders outside the moderation queue, so there's no conflict.

Closes #2850

### DIFF
--- a/app/user/moderation/UserModerationTab.js
+++ b/app/user/moderation/UserModerationTab.js
@@ -332,7 +332,7 @@ const UserModerationQueue = ({
     token.startsWith('~') || token.includes('@') || isValidDomain(token.toLowerCase())
 
   const getProfiles = async () => {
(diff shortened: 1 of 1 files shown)


xkopenreview:
i do feel it's a bit faster

xkopenreview (approved):

PR #2856, opened 2026-07-22
Fix/ ProfileTag - update label rendering rules for improved clarity

xkopenreview:
currently when tag label is 
>user sent document

it will be rendered in green color

sometimes the email user used to send documents are different from the one in profile
so the tag becomes
>user sent document from email.domain

this pr should update the logic to include this case


xkopenreview:
given the changes in upload workflow maybe this is not relevant anymore

melisabok (approved):

PR #2858, opened 2026-07-23
Fix/ Invalid Array Length error

xkopenreview:
the error happens in edge browser when loading all reviewer profiles in reviewers group
the count returned in the call is greater than the actual profile objects count by 1

causing the check of whether to get subsequent results to fail (totalCount - initialResults.length = 1)
```javascript
if (totalCount - initialResults.length <= 0) return initialResults
``` 

so the offset array is generated using negative value causing the error

this pr should mitigates the error but does not fix the root cause


xkopenreview:
the profiles causing the issue have been found
maybe this pr is not required anymore

PR #2871, opened 2026-08-06
Fix/ Contact form - add new contact option to appeal for rejection

xkopenreview:
this pr should modify the contact form:
1. remove institutional related code which was commented out
2. add a new topic option to appeal for profile rejection which include a profile ID input but is not mandatory in case some user can't find profile id during registration


app/contact/page.js
```diff
@@ -30,16 +29,16 @@ export default function Page() {
   const organizationSubject = 'A conference I organized'
   const committeeSubject = 'I am a reviewer or committee member'
   const createProfileSubject = 'I am trying to create my profile'
+  const profileRejectedSubject = 'My profile was rejected and I want to appeal'
```
melisabok: We do not mention the word "rejected" in the moderation email so I would prefer to not put it here. 

"My profile requires additional verification and I do not have it..." or something like that @deanaOR for feedback

PR #2882, opened 2026-08-10
Deploy production web via blue/green cutover

tyler-dominguez:
## Summary

The production web MIG is now flexible (mixed machine families for stockout
resilience), which forces an `OPPORTUNISTIC` update policy and makes an in-place
rolling update impossible — `rolling-action start-update` (the current deploy step)
errors on a flexible MIG. This switches the web rollout to a blue/green cutover.

## Changes

Replaces the `terraform apply` + `rolling-action start-update` + `wait-until` steps
with a blue/green flow (build, Packer, and image→tag mapping are unchanged):

- Determine the active color and its instance count from the load balancer.
- Warm the idle color to that count on the new image (active color untouched, still
  serving 100%).
- Wait for the idle color to reach that count healthy on the LB before flipping.
- Flip `capacity_scaler` on both web backend services — the external GLB and the
  internal ILB (nginx path) — in one apply.
- Settle the new active color (floor + prewarm) and drain the old one to 0.
- On any failure before the flip, the previous color keeps serving and the idle
  color is drained.

`terraform` now runs on the rollback path too, since a rollback is the same cutover
onto the mapped image. Dispatch inputs (`tag`, `force_rebuild`) are unchanged.

## Depends on

openreview/openreview-cloud#117 — the blue/green Terraform for the web MIGs and the
backend-service flip. Must be merged before this workflow will run correctly.

## Rollback

Unchanged for operators: re-run the workflow with the previous `tag`; it reuses that
tag's baked image and cuts over to it (no rebuild).

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -306,15 +306,6 @@ jobs:
             -var="green_image=${{ steps.state.outputs.green_img }}" \
             -var="warm_idle_size=0"
 
-          OLD_GROUP=${{ steps.state.outputs.active_group }}
-          insts=$(gcloud compute instance-groups managed list-instances "$OLD_GROUP" \
-            --region=us-central1 --project=$PROJECT_ID --format="value(instance.basename())" | paste -sd, -)
-          if [ -n "$insts" ]; then
-            echo "Draining old color $OLD_GROUP: $insts"
-            gcloud compute instance-groups managed delete-instances "$OLD_GROUP" \
-              --instances="$insts" --region=us-central1 --project=$PROJECT_ID
-          fi
-
       - name: Failover to previous color if the cutover failed before flipping
         if: failure() && steps.flip.outcome != 'success'
         env:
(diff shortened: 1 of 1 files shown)


xkopenreview:
is it possible that the idle group are not really idling when making 2 consecutive deployments?

tyler-dominguez:
> is it possible that the idle group are not really idling when making 2 consecutive deployments?

Yes, good catch. Added guardrails for this and changed how instances get marked for deletion to reduce period of time this can occur.

.github/workflows/deploy-prod.yml
```diff
@@ -178,43 +177,166 @@ jobs:
...
+          N=${{ steps.state.outputs.n }}
+          deadline=$(( $(date +%s) + 900 ))
+          while :; do
+            healthy=$(gcloud compute backend-services get-health $GLB --global --project=$PROJECT_ID \
+              --format=json | jq --arg g "$IDLE_GROUP" \
+              '[.[] | select(.backend|test("/"+$g+"$")) | .status.healthStatus[]? | select(.healthState=="HEALTHY")] | length')
+            healthy=${healthy:-0}
+            echo "  $IDLE_GROUP: $healthy/$N healthy"
+            [ "$healthy" -ge "$N" ] && break
+            [ "$(date +%s)" -ge "$deadline" ] && { echo "TIMEOUT: idle did not reach $N healthy"; exit 1; }
```
tyler-dominguez, line 279: @xkopenreview

carlosmondra (approved):

melisabok (approved):

PR #2890, opened 2026-08-13
Fix/ ProfileEditor History - Show section error and row error

xkopenreview:
currently career/education history section shows error in a tooltip inside each input

this pr should change it to 
- show section wide errors which are not related to specific career/education history row on top 
- show all errors about a row below the row

### DIFF
--- a/ThemeProvider.js
+++ b/ThemeProvider.js
@@ -4,6 +4,8 @@ import { ConfigProvider } from 'antd'
 const primaryColor = '#3e6775'
 const backgroundWhite = '#fffdfa'
 const orRed = '#8c1b13'
+const orRedLight = '#ee8a83'
+const alertRed = '#f2dede'
 const subtleGray = '#616161'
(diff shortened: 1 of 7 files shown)


melisabok (approved):

emilygrabowski:
1) I was able to save my profile with a position with no start date if I also include no end date - not sure if this is expected or not. 
2) When I put in an invalid end year (0), the error showed at the top and other errors in the section didn't show -- same for start date
3) This error could be clearer:  he is not a valid domain. Domains should not contain "http", "www", or and special characters like "?" or "/". -->  he is not a valid domain. Domains should be in the format of the email domain for your institution (e.g. umass.edu)

xkopenreview:
@emilygrabowski 
1 is expected
can you add a screenshot for 2 and 3?

emilygrabowski:
Screenshot 2026-08-18 at 2 25 07 PM

Screenshot 2026-08-18 at 2 26 13 PM

PR #2891, opened 2026-08-14
Add "Leave Impersonation" to the user menu

melisabok:
## Motivation

Once impersonating a user there was no way back to your own session from the UI: the nav shows "(Impersonated)" but the only exits were logging out entirely or manually clearing cookies. Support workflows often require impersonating several users in a row (e.g. checking a venue's consoles as different reviewers), so leaving impersonation should be one click and should drop you where the next impersonation starts.

### New `app/(Home)/LeaveImpersonationLink.js`

Mirrors the existing `LogoutLink` pattern: posts to the new API endpoint `POST /impersonate/exit`, which invalidates the impersonated access/refresh tokens server-side and restores the impersonator's own session cookies. On success it navigates (full page load, `window.location.assign`) to `/user/impersonate`, so the app re-reads the restored cookies and the impersonator lands on the Impersonate User page — with their Previous Impersonations list handy for jumping to the next user. Errors surface via `promptError`, matching `LogoutLink`.

### Nav menus (`NavLg`, `NavMd`, `NavSm`)

The item is rendered only while impersonating (`user.impersonator` is set — the same flag that already drives the "(Impersonated)" label):

- **NavLg / NavMd** (antd Dropdown): added between the divider and Logout.
- **NavSm** (mobile drawer): added directly above Logout; closes the drawer on tap.

No change when not impersonating — the menus render exactly as before.

## Testing

- Verified end to end against a local stack (test data from `test_reviewers_only.py`) with headless Firefox:
  - Impersonating `~ReviewerOne_ABCD1` as the super user, the desktop dropdown shows Profile / Password & Security / **Leave Impersonation** / Logout.
  - Clicking it lands on `/user/impersonate` with the nav showing the original user again (no "(Impersonated)").
  - Not impersonating: the item is absent.
- `oxlint` and `oxfmt --check` clean on the four touched files.

## Dependencies

Requires the openreview-api branch `worktree-fix+1199-allow-impersonation-mode-to-be-logged-out`, which adds `POST /impersonate/exit` (blacklists the abandoned impersonated tokens, restores the impersonator's cookies). Merge/deploy the API side first — without it the link fails with a 404, surfaced via `promptError`.

### DIFF
--- a/app/(Home)/LeaveImpersonationLink.js
+++ b/app/(Home)/LeaveImpersonationLink.js
@@ -1,0 +1,33 @@
+'use client'
+
+import Link from 'next/link'
+import api from '../../lib/api-client'
+
+export default function LeaveImpersonationLink({ className, style, onClick }) {
+  const handleLeaveImpersonation = async (e) => {
+    e.preventDefault()
+    onClick?.()
+
+    try {
+      await api.post('/impersonate/exit')
+      window.location.assign('/user/impersonate')
+    } catch (error) {
+      // oxlint-disable-next-line no-console
+      console.log('Error in LeaveImpersonationLink', {
+        page: 'Home',
+        component: 'LeaveImpersonationLink',
+        apiError: error,
+        apiRequest: {
+          endpoint: '/impersonate/exit',
+        },
+      })
+      promptError(error.message)
+    }
+  }
+
+  return (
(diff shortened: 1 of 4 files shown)


melisabok:
Screenshot 2026-08-14 at 9 50 59 AM

xkopenreview (approved):

PR #2895, opened 2026-08-19
Show the human verification challenge when an edge post is blocked

melisabok:
## Motivation

The API now enforces `humanVerificationRequired` on edges and tags, and the Bid and Expertise
Selection invitations set it. Nothing in the bidding or expertise consoles knew about that: a reviewer
who crossed the limit got a raw `HumanVerificationRequiredError` toast from `promptError` and no way
to continue bidding.

Note editing, file upload, signup, recruitment and the delete-note modal already handle this — each
calls `useTurnstileToken` and renders the widget next to its own submit control. The edge consoles are
the ones that were never wired up. A second gap: the Human Verification field on `/invitation/edit`
was hidden for edge and tag invitations, so the config could be read but never edited.

### New `hooks/useHumanVerification.js`

The retry glue the existing consumers each hand-roll, factored out because six call sites needed it.
It wraps a request so a `HumanVerificationRequiredError` **holds** the request instead of failing it:
the promise handed back to the caller stays pending, the Turnstile widget renders, and once the user
solves it the request is replayed with the `cf-turnstile-token` header and the caller's promise settles
with the replay. Call sites only merge the options they are given into their `api.post`:

```js
const result = await runWithHumanVerification((requestOptions) =>
  api.post('/edges', getBidObjectToPost(...), requestOptions)
)
```

Only one request can be held at a time; if a second is blocked while one is pending, the earlier one is
rejected so whatever is waiting on it resets rather than hanging. The widget is scrolled into view when
it appears, since in a long bid list it is nowhere near the row the user just clicked.

### New `components/HumanVerificationChallenge.js`

The explanation line plus the same bare `<div ref={turnstileContainerRef} />` the other consumers
render. No modal — the container is always mounted and `useTurnstileToken` fills it only while a
request is waiting.

### Consoles

Wired into all six edge-posting sites, at the top of each tab so the challenge is reachable regardless
of how far down the list the user was:

- `components/webfield/BidConsole.js` — `AllSubmissionsTab`, `NoBidTab`, `BidOptionTab`
- `components/webfield/ProfileBidConsole.js` — `AllSubmissionsTab`, `BidOptionTab` (SAC-bids-on-AC)
- `components/ExpertiseSelector.js` — `toggleEdge`

### `components/invitation/InvitationGeneral.js`

`const isNoteInvitation = invitation.edit?.note` gated the Human Verification editor in three places —
initial form state, save payload, and JSX — so the row never rendered for an edge or tag invitation,
while the read-only view (which has no gate) still showed the value. The gate predates the API
supporting edges and tags; `humanVerificationRequired` sits at the top level of the invitation schema
regardless of kind, and the check is wired into `models/edit.js`, `models/edge.js` and `models/tag.js`.
Removed it, along with the now-unused flag.

Also, clearing the field did nothing: the value was only included when non-empty, so a config could be
added but never removed. It now sends `{ delete: true }` when the box is empty, matching how
`maxReplies` / `minReplies` / the dates already behave.

## Testing

- Two tests added to `unitTests/BidConsole.test.js`, mirroring the existing FileUploadWidget ones:
  the challenge and its explanation appear when the bid post returns `HumanVerificationRequiredError`,
  and the bid is replayed with `cf-turnstile-token` once the token arrives.
- `npx jest` — 769 passed, 2 skipped, 80 suites.
- `oxlint` and `oxfmt` clean on the touched files.

Note the unit tests click the radio `<input>` directly rather than the `<label>`. Clicking the label
fires `updateBidOption` twice — label activation dispatches a second click that bubbles back to the
same handler in `BidWidget.js` — so every bid currently posts its edge twice, which also halves the
effective rate limit. Pre-existing and left alone here, but worth a separate fix.

## Dependencies

- openreview-api branch `worktree-fix-human-verification-edges-tags` — enforces the check on edges and
  tags and returns `HumanVerificationRequiredError`.
- openreview-py — sets `DEFAULT_EDGE_TAG_HUMAN_VERIFICATION` (100/hour) on the Bid and Expertise
  Selection invitations.

Safe to merge ahead of either: with no invitation carrying the config, nothing ever raises the error
and the challenge never renders.


melisabok:
closing, no needed.

PR #2896, opened 2026-08-19
Fix/ Moderation - show journal requests

xkopenreview:
this pr should show update moderation page
- venue requests
- deployed venues

to include journal requests
logic is same as workflow requests

### DIFF
--- a/app/user/moderation/(VenueRequests)/NewVenueRequestCount.js
+++ b/app/user/moderation/(VenueRequests)/NewVenueRequestCount.js
@@ -16,13 +16,12 @@ export default function NewVenueRequestCount({ children }) {
           },
           {
             invitation: `${process.env.SUPER_USER}/Support/Venue_Request/-/Conference_Review_Workflow`,
-            select: `parentInvitations,content.venue_id`,
+            select: 'content.venue_id',
           },
           { includeVersion: true }
         )
         .then((response) =>
           response?.notes?.filter((p) => {
-            if (p.parentInvitations) return false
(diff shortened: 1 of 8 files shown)


app/user/moderation/(VenueRequests)/NewVenueRequestCount.js
```diff
@@ -29,7 +29,18 @@ export default function NewVenueRequestCount({ children }) {
...
         )
 
-      setPendingVenueRequestCount(undeployedVenueRequests?.length)
+      const undeployedJournalRequests = await api
+        .get('/notes', {
+          invitation: `${process.env.SUPER_USER}/Support/-/Journal_Request`,
+          select: `parentInvitations,content.venue_id`,
+        })
+        .then((response) =>
+          response?.notes?.filter((p) => !p.parentInvitations && !p.content?.venue_id?.value)
```
melisabok: why do you need parentInvitations here?

melisabok: is this to check if the venue is deployed or not? checking venue_id existence should be enough, right?

xkopenreview: parent invitations was added when venue request notes are queried using regex Support/Venue_Request.*
so is useless now. removed in [3e130b2](https://github.com/openreview/openreview-web/pull/2896/commits/3e130b21ce0eb7acc6515404ab1dc533e3ec0371)

celestemartinez (approved):

PR #2902, opened 2026-08-24
Fix/ Submissions - handle clearance token

xkopenreview:
currently submissions page will fail for guest due to missing clearance token handling

this pr should fix the issue


carlosmondra (approved):

xkopenreview:
requires #2836 to pass

PR #2904, opened 2026-08-25
Feat/deploy dev via cloud run

tyler-dominguez:
## Migrate dev web deployment to Cloud Run

Repoints the two dev-deploy paths from building the Next.js app on the dev VM to building + deploying the `openreview-web-dev` Cloud Run service via Cloud Build. Same auth maintained.

The build config and Dockerfile live in `openreview-cloud` (`Terraform/dev/web-cloudrun/build`), which each job checks out via a GitHub App token (mirroring `deploy-prod.yml`), then runs `gcloud builds submit`. Cloud Build clones this repo at the target ref, bakes `staging.env`, pushes the image, and `gcloud run deploy`s it.

### Changes
- **`deploy-dev.yml` (manual Staging Deployment):** replaces the per-VM SSH `build-web.sh`/`deploy-web.sh` loop with `gcloud builds submit` using the dispatched `tag` (`_WEB_REF=${TAG}`).
- **`build.yml` (deploy job, on push to master):** same swap with `_WEB_REF=master`. The `lint`/`build` CI jobs are unchanged. The `deploy-openreview-py.sh` step is retained as an SSH step — it refreshes openreview-py on the dev box itself and is unrelated to web?

### Tested staging deployment
https://github.com/openreview/openreview-web/actions/runs/33192132067

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,6 +227,20 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ secrets.APP_CLIENT_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: openreview
+          repositories: openreview-cloud
+      - name: Checkout openreview-cloud
+        uses: actions/checkout@v6
+        with:
+          repository: openreview/openreview-cloud
+          path: openreview-cloud
+          token: ${{ steps.app-token.outputs.token }}
       - name: Authenticate with Google Cloud
         id: auth
         uses: google-github-actions/auth@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -223,6 +223,10 @@ jobs:
 
     needs: build
 
+    concurrency:
+      group: dev-web-deploy
+      cancel-in-progress: false
+
     if: github.ref == 'refs/heads/master' && (github.event_name != 'repository_dispatch' || github.event.action != 'openreview-api-updated')
```
tyler-dominguez, line 230: @xkopenreview Do we want to stop deploying to web dev on this condition? I checked and saw runs completing deployments from the Next.js CI step

melisabok (approved):

.github/workflows/build.yml
xkopenreview: which condition?
i think the condition is to exclude run triggered by api and py changes right?

tyler-dominguez: Yes the exclusion is good, but do we want the case where it is true ever to happen? When it is a push to master and it isn't an api update dispatch? Because this step is being reached.

xkopenreview: yes. when it's a commit to master from web repo then the updated master should be deployed to dev

PR #2905, opened 2026-08-25
Fix/ Forum Institution Authors - consolidate institution domains

xkopenreview:
currently the same affiliated institution might be shown multiple times when the domain is different

this pr should fix this issue by querying institution objects and look up the domain so that different domains which belong to the same institution object are consolidated

when there are 20+ current institution domains (many authors or many active institutions), the look up is skipped to avoid constructing too long a query


melisabok (approved):

celestemartinez (approved):
Tested this, works well!

PR #2906, opened 2026-08-26
Fix/ Profile edit - editable dob

xkopenreview:
this pr should make birth date always editable according to the discussion in sync meeting
this pr should also remove warning text that birth date is not editable in profile edit page and sign up page

### DIFF
--- a/app/signup/Signup.js
+++ b/app/signup/Signup.js
@@ -197,11 +197,6 @@ const SignupForm = ({ setSignupConfirmation }) => {
                 style={{ flex: '1.5 1 0' }}
               />
             </Flex>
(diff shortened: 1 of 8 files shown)


xkopenreview:
this pr should also fix the date drifting issue caused by improper fix applied in #2829
date.tz('UTC', true) mutates its receiver for offset-0 zones instead of creating new date object
while antd date picker still holds the date obj for display which cause (in display) one day drift for UTC+ users

the valued saved is correct so some users notice the issue is fixed by itself

carlosmondra (approved):

melisabok (approved):

PR #2922, opened 2026-09-02
Fix/ Allow parental consent to be deleted

xkopenreview:
this pr should fix the request to be able to delete parental consent which was not deletable

this pr should also close preview when there are no documents for preview


xkopenreview:
@deanaOR parental consent can be delete in preview
if the upload link is not expired, the user can use the same link to continue uploading

melisabok (approved):

PR #2929, opened 2026-09-09
Fix/ Update venue organizer agreement text

xkopenreview:
venue_organizer_agreement text is updated
so update in setup so tests pass


xkopenreview:
@celestemartinez this is required for tests to pass

celestemartinez (approved):
Thanks!

PR #2938, opened 2026-09-16
AC/PC/SAC consoles: render custom stage replies to reviews next to their review

melisabok:
## Summary

The consoles' `customStageInvitations` setting previously rendered a single custom stage reply per submission in the Meta Review Status column. This PR extends it to support custom stages that reply to reviews (e.g. an `AI_Review_Detection` stage posting one reply per official review, as used by ICLR): all matching replies are now collected and split by what they reply to, and replies to a review are rendered in the Official Review Progress column directly under the review they reply to — in the AC, PC and SAC consoles — so each review's AI detection result (or any other per-review signal) shows next to the review itself.

Example webfield config on the console group:

```js
customStageInvitations: [{
  name: 'AI_Review_Detection',
  displayField: 'label',
  extraDisplayFields: ['score', 'link']
}],
```

### `components/webfield/AreaChairConsole.js`
- The custom stage replies of each configured stage are split into two collections by their `replyto`:
  - **`customStageReviewReplies`** — replies to an official review; stored at the row level and rendered in the review status column.
  - **`customStageMetaReviewReplies`** — replies to a meta review or to the forum (the default for custom stages, e.g. a second review invitation or a meta review confirmation); stored in `metaReviewData` and rendered in the Meta Review Status column, as before.
- All matching replies are collected (`filter`) instead of only the first one (`find`).
- Fixed a latent bug where a reply was matched against every configured custom stage id instead of the stage being processed.
- Replies rendered in the meta review column are labeled with the last segment of the invitation id prefix (`invitationId.split('/-/')[0]`) when they are not forum-level replies, e.g. `Meta Review Confirmation (Meta Review1)`, to distinguish multiple replies.

### `components/webfield/NoteReviewStatus.js`
- New `CustomStageReviewReplies` component: renders each custom stage reply under the review it replies to (matched by `replyto`), showing the stage name, the display field, the extra display fields and a link to read the reply in the forum. Values starting with `https` are rendered as links.
- `AcPcConsoleNoteReviewStatus` accepts an optional `customStageReviewReplies` prop and forwards it to both reviewer row variants. It is shared by the AC, PC and SAC consoles, so all three render the same blocks.

### `components/webfield/NoteMetaReviewStatus.js`
- `AreaChairConsoleNoteMetaReviewStatus` renders `metaReviewData.customStageMetaReviewReplies` (all entries, with the invitation prefix label).

### `components/webfield/ProgramChairConsole.js` and `components/webfield/SeniorAreaChairConsole.js`
- Both consoles build a row-level `customStageReviewReplies` collection (replies to an official review, per stage) and exclude those replies from their existing single-object `customStageReviews`, so nothing renders twice. The PC console Overview tab statistics (`repliesPerSubmission`) and the `metaReviewAgreement` logic are unchanged.

### `components/webfield/ProgramChairConsole/PaperStatus.js` and `components/webfield/SeniorAreaChairConsole/PaperStatus.js`
- Pass the review replies into the shared `AcPcConsoleNoteReviewStatus`.

### `components/webfield/ProgramChairConsole/PaperStatusMenuBar.js` (shared by PC and SAC)
- The custom stage export column and sort options join the values across the review replies when present and fall back to the existing single-reply behavior for forum/meta review stages.

### `components/webfield/AreaChairConsoleMenuBar.js`
- The custom stage search/sort options read from both collections and join the values across all replies.
- A stage without `extraDisplayFields` no longer loses its display field sort option.

## Backward compatibility

- Custom stages replying to the forum (e.g. a second review invitation configured through `customStageInvitations`, or `Meta_Review_Confirmation` posted as a forum reply) keep rendering in the Meta Review Status column as before, in all three consoles.
- Venues that show multiple replies per stage now see all of them instead of only the first one found.
- The PC/SAC single-object `customStageReviews` model (used by string-path query search, the Overview statistics and export fallbacks) is preserved.

## Venue configuration notes

- The replies only render for users who can read the notes: the custom stage `readers` must include the console's role (e.g. `SENIOR_AREA_CHAIRS_ASSIGNED` for the replies to show in the SAC console).
- The SAC console can only find its assignments when the assigned SACs can read their `Submission<number>/Senior_Area_Chairs` group (`senior_area_chair_identity_readers`); the companion openreview-py PR defaults this for template venue deployments.

## Tests

- `unitTests/AreaChairConsole.test.js`: new test mounting the console with replies to reviews (`AI_Review_Detection`), to the forum (`Second_Round_Review`, a second review invitation configured as a custom stage) and to the meta review (`Meta_Review_Confirmation`), asserting each entry is routed to the right column with the right source label.
- `unitTests/NoteReviewStatus.test.js` (new): asserts each reply renders under the review it replies to, `https` values render as links, and reviewer rows are unchanged when no replies are passed.
- `unitTests/AreaChairConsoleMenuBar.test.js`: new test asserting the custom stage sort options read from both collections and handle rows without replies.
- `unitTests/PaperStatusMenuBar.test.js`: new test asserting the PC/SAC sort options join review replies and fall back to the single-reply meta review data.
- Full unit suite passes: 84 suites, 796+ tests.
- Verified end to end against a local stack running `test_iclr_conference_with_templates.py` (openreview-py), checking the rendering in the browser in all three consoles: as the assigned AC, as the assigned SAC and as a PC.

Companion openreview-py PR:
- Enables `customStageInvitations` on the AC/SAC/PC group webs in the ICLR templates test, and adds `SENIOR_AREA_CHAIRS_ASSIGNED` to the AI review detection stage readers so the SAC console can display the replies.
- Defaults template venue deployment so assigned SACs/ACs can see reviewer identities (`reviewer_identity_readers`) and assigned SACs can read their own submission group (`senior_area_chair_identity_readers`) — without the latter the SAC console cannot find its assignments.


enrubio (approved):
Checked the PC/SAC/AC consoles and it looks great!

xkopenreview:
i think jsdoc comments of related consoles should be updated with example
so it can be converted to openreview doc pr

components/webfield/ProgramChairConsole/PaperStatusMenuBar.js
xkopenreview, line 106: the search path is not updated
so when performing query search it will still look for metaReviewData.customStageReviews instead of custom stage review replies

components/webfield/ProgramChairConsole/PaperStatusMenuBar.js
```diff
@@ -456,8 +460,13 @@ const PaperStatusMenuBar = ({
...
                   getValue: (p) =>
-                    p.metaReviewData?.customStageReviews?.[camelCase(invitation.name)]
-                      ?.content?.[extraDisplayField]?.value ?? 'N/A',
+                    p.customStageReviewReplies?.[camelCase(invitation.name)]
+                      ?.map((q) => q.content?.[extraDisplayField]?.value)
+                      ?.filter((q) => q !== undefined && q !== null)
+                      ?.join(' ') ||
+                    (p.metaReviewData?.customStageReviews?.[camelCase(invitation.name)]
+                      ?.content?.[extraDisplayField]?.value ??
+                      'N/A'),
```
xkopenreview, line 469: is it a valid usecase to have both customstage review reply and custom stage review?
customStageReviews will be dropped by ||

melisabok: we can have both, and the custom stage review will keep displayed in the metareview column

PR #2940, opened 2026-09-17
Fix/ Profile Edits

xkopenreview:


melisabok:
it would be nice to show the `tauthor` of each edit so we can track who is moderating each profile:

Screenshot 2026-09-22 at 5 02 51 PM

melisabok:
This is how it looks the edits in the profile page using my branch, I like the icon next to the text and using a color in the font instead of background color:



Screenshot 2026-09-22 at 9 44 18 PM
Screenshot 2026-09-22 at 9 44 02 PM

melisabok:
I think the forms must be rendered next to the document and avoid showing the profile info. The form should be prefilled with existing data so the support user can easily compare the values against the document. 

The text area must be the latest field to show not the first one.

melisabok:
Relations rendering:

Screenshot 2026-09-22 at 9 57 18 PM

xkopenreview:
@melisabok please take a look again.
slightly modified your design

PR #2943, opened 2026-09-18
Workflow timeline: load process logs by invitation id instead of by venue prefix

melisabok:
## Motivation

The workflow timeline loaded the process logs of the whole venue:

```js
api.getAll(
  '/logs/process',
  { invitation: `${groupId}.*`, select: 'id,sdate,edate,invitation,status,log' },
  { resultsKey: 'logs' }
)
```

That regex matches every log of the venue, including one per submission for each invitation with a
process function — `ICLR.cc/2027/Conference/Submission1234/-/Official_Review` and the like. ICLR 2027
has 40k submissions, so the timeline pages hundreds of thousands of logs into the browser on every
load, and again after every workflow event.

The timeline never shows any of them. It only renders a log status for the invitations that have date
process functions (`EditInvitationProcessLogStatus` is rendered when `invitation.dateprocesses?.length
> 0`), and it only reads two things per invitation: whether there is a running log, and the last
finished one.

### `components/group/WorkflowInvitations.js`

- `loadProcessLogs` now takes the invitation ids to load and queries them one at a time by exact id
  (`{ invitation: <id> }`), instead of one paged prefix query for the venue. The exact id is also what
  the `{ invitation: 1, status: 1 }` index on the logs collection is built for.
- The ids are the invitations shown in the timeline that have date process functions, collected right
  after the invitations load and kept in `processLogInvitationIdsRef`:

  ```js
  processLogInvitationIdsRef.current = filteredInvitations.flatMap((p) =>
    p.dateprocesses?.length > 0 ? p.id : []
  )
  ```

  This means the logs can no longer be fetched in the same `Promise.all` as the invitations — they are
  loaded right after `filterWorkflowInvitations`, before the invitations are formatted, so
  `formatWorkflowInvitation` still gets the logs it needs for the section class.
- The results keep the existing client-side `orderBy(logs, ['edate'], ['desc'])`. A running log has no
  `edate` yet and lodash sorts `undefined` first in a `desc` sort, so a running log still takes
  precedence over the last finished log of the same invitation, exactly as before.
- The workflow event handler no longer reloads everything. The `date-process-updated` payload carries
  the `invitation` the log belongs to, so the ids reported since the last refresh accumulate in
  `pendingLogInvitationIdsRef` and only those are re-queried when the 5s debounce fires. Events for
  invitations the timeline does not track (the per-submission ones) are ignored, and the logs of the
  other invitations are kept as they are.

## Behavior notes

- `getSectionClass` marks a non-stage invitation `section-active` when it has a running log. It now
  only sees the logs of the invitations with date process functions, so a timeline invitation without
  them no longer turns active because of an unrelated edit process log.
- Each query returns that invitation's logs up to the API's default page limit (1000) rather than a
  single row. For a date process invitation that is normally a handful of logs, but an invitation that
  also has an edit process function — `/-/Submission` gets one log per submission — still returns up to
  1000 rows. Cutting that down needs a `limit` on the query plus a way to ask the API for date process
  logs only: a running log has no `edate`, so it sorts last server side and a plain `limit` would drop
  it.

## Tests

- `unitTests/WorkflowInvitations.test.js` (new):
  - the logs are queried once per invitation with date processes, by exact id, and never through a
    prefix `getAll`;
  - a running log is shown instead of the last finished log of the same invitation;
  - a workflow event refreshes only the invitation it reports;
  - a workflow event for an invitation that is not in the timeline is ignored.
- All four fail against the previous implementation.
- Full unit suite passes: 84 suites, 796 passed, 2 skipped.


components/group/WorkflowInvitations.js
```diff
@@ -1086,8 +1115,15 @@ const WorkFlowInvitations = ({ group }) => {
 
   useEffect(() => {
     if (!events) return
+    const eventInvitationId = events.data?.invitation
```
xkopenreview: i think api will always send invitation

components/group/WorkflowInvitations.js
xkopenreview, line 1065: the logs here is only used to check running status of of run invitations
i think maybe it's fine to show it inaccurately at first load? so that the wait on loadProcessLogs can be dropped

components/group/WorkflowInvitations.js
```diff
@@ -674,6 +674,10 @@ const WorkFlowInvitations = ({ group }) => {
   const [missingValueInvitationIds, setMissingValueInvitationIds] = useState([])
   const events = useSocket('venue/workflow', ['date-process-updated'], { venueid: groupId })
   const workflowInvitationsRef = useRef({})
+  // Ids of the invitations the process logs are loaded for, and the ids reported by the
+  // workflow events since the last refresh.
+  const processLogInvitationIdsRef = useRef([])
+  const pendingLogInvitationIdsRef = useRef(new Set())
```
xkopenreview: i would considering renaming the refs
it's a bit confusing with the comments

xkopenreview: some rename ideas ai suggested
processLogInvitationIdsRef => invitationIdsWithLogsRef 
pendingLogInvitationIdsRef => pendingRefreshIdsRef

components/group/WorkflowInvitations.js
xkopenreview: discussed this with Melisa
better to show accurate active status of run invitations at first load
also with changes in this pr there won't be many invitations to load so drop this

melisabok:
@xkopenreview all the feedback has been addressed

xkopenreview (approved):

